PR #11, opened 2015-03-02
For #28028, update desktop to use core login framework

jfboismenu-adsk:
- Uses tank.platform.interactive_authentication module for managing login and logout.
- Uses tank.util.login.get_current_user to get the Human User that is logged in.
  - get_current_user code introduces a regression where the Shotgun account associated to the os user
   will be always be returned insead of the one associated with the login credentials. Once the core
   is updated so that get_current_user returns the user currently authenticated as, this regression will
   disappear.

### DIFF
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -93,7 +93,7 @@ class DesktopWindow(SystrayWindow):
         self.ui.apps_button.style().unpolish(self.ui.apps_button)
         self.ui.apps_button.style().polish(self.ui.apps_button)
 
-        interactive_authentication.ui_login()
+        interactive_authentication.ui_authenticate()
         connection = shotgun.get_sg_connection()
 
         engine = sgtk.platform.current_engine()
@@ -105,7 +105,7 @@ class DesktopWindow(SystrayWindow):
 
         # User menu
         ###########################
-        current_user = login.get_current_user()
+        current_user = self._get_current_user()
         thumbnail_url = current_user.get("image")
         if thumbnail_url is not None:
             (_, thumbnail_file) = tempfile.mkstemp(suffix=".jpg")
@@ -503,8 +503,11 @@ class DesktopWindow(SystrayWindow):
         self.update_project_config_widget.show()
         self.project_overlay.hide()
 
+    def _get_current_user(self):
+        return login.get_current_user(sgtk.platform.current_engine().tank)
+
     def __populate_pipeline_configurations_menu(self, pipeline_configurations, selected):
-        user = login.get_current_user()
+        user = self._get_current_user()
 
         primary_pc = None
         extra_pcs = []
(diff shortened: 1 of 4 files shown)


jfboismenu-adsk:
This should be reviewed in tandem with

python/tk_desktop/desktop_window.py
```diff
@@ -504,9 +503,11 @@ def show_update_project_config(self):
         self.update_project_config_widget.show()
         self.project_overlay.hide()
 
+    def _get_current_user(self):
+        return login.get_current_user(sgtk.platform.current_engine().tank)
```
manneohrstrom: Is this inside the engine? Could you use `sgtk.platform.current_bundle()`? That's the standard way to get the app/engine object from code inside the app/engine (but only works if you have used `import_module`).

manneohrstrom: `get_current_user()` tries to match the operating system user right? Should we add a comment here that this is going to be fixed soon? Or is it ok this way?

python/tk_desktop/project_commands_model.py
```diff
@@ -267,7 +264,7 @@ def _handle_command_triggered(self, item, command_name=None, button_name=None,
             "event_type": self.APP_LAUNCH_EVENT_TYPE,
             "project": self.__project,
             "meta": {"name": command_name, "group": group_name},
-            "user": login,
+            "user": login.get_current_user(engine.tank),
```
manneohrstrom: sgtk please!

python/tk_desktop/project_model.py
```diff
@@ -218,10 +217,7 @@ def __init__(self, parent, overlay_parent_widget):
         """ Constructor """
         ShotgunModel.__init__(self, parent, download_thumbs=True)
 
-        # override the default connection to use one that is tied to the current login
-        connection = ShotgunLogin.get_instance_for_namespace("tk-desktop").get_connection()
-
-        self.set_shotgun_connection(connection)
+        self.set_shotgun_connection(shotgun.create_sg_connection())
```
manneohrstrom: does this need its own instance? Or can it grab a shared instance?

manneohrstrom:
CR comments: 
I wasn't sure about the use of the current login - are we making that chance in anticipation of that coming change? Also the use of current engine perhaps feels little non-standard (but should work fine!) - wondering if current-bundle may be better?

python/tk_desktop/desktop_window.py
jfboismenu-adsk: I attached a comment to the commit to that effect, but it should be in the code as well. Will do.

jfboismenu-adsk: This code is part of the engine, but I don't understand what's the difference between calling the two? current_engine has the merit of being explicit about what I think I am getting back.

python/tk_desktop/project_model.py
jfboismenu-adsk: It should probably reuse a thread-local one. I'll grab the sgtk.shotgun instance.

PR #13, opened 2015-03-04
For #26554, better restart logic when core is updated

jfboismenu-adsk:
About Box now displays Core version.

### DIFF
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -859,5 +859,9 @@ class DesktopWindow(SystrayWindow):
                 Engine Version %s<br/>
                 Core Version %s
             </center>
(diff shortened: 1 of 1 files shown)


python/tk_desktop/desktop_window.py
```diff
@@ -855,7 +856,8 @@ def handle_about(self):
         about = AboutScreen(parent=self, body="""
             <center>
                 App Version %s<br/>
-                Engine Version %s
+                Engine Version %s<br/>
+                Core Version %s
             </center>
-        """ % (engine.app_version, engine.version))
+        """ % (engine.app_version, engine.version, pipelineconfig_utils.get_currently_running_api_version()))
```
robblau: Line length, this should be broken up.

PR #14, opened 2015-03-09
Displays error correctly (and potential fix) in project setup error

mathieurivest:
When the error is due to a tank name already existing without a pipelineconfig present.

### DIFF
--- a/python/tk_desktop/setup_project.py
+++ b/python/tk_desktop/setup_project.py
@@ -51,6 +51,15 @@ class SetupProject(QtGui.QWidget):
             ret = setup.exec_()
             self.setup_finished.emit(ret == setup.Accepted)
 
+        except TankErrorProjectIsSetup, e:
+            from error_dialog import ErrorDialog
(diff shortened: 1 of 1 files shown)


python/tk_desktop/setup_project.py
```diff
@@ -46,11 +48,18 @@ def do_setup(self):
             if is_on_top: 
                 self._set_top_window_on_top(False)
 
-
             setup = adminui.SetupProjectWizard(self.project, self)
             ret = setup.exec_()
             self.setup_finished.emit(ret == setup.Accepted)
 
+        except TankErrorProjectIsSetup, e:
+            error_dialog = ErrorDialog("Toolkit Setup Error",
```
jfboismenu-adsk: A bit unfair to ask this during your code review since you weren't the one who introduced the limitation, but why don't we force the project creation like we do with tank?

robblau: :)
The plan is to bring up a dialog that will prompt for the force, but yup... in retrospect that would be a better way to go.

PR #16, opened 2015-04-02
For #28769, update desktop view so template project's do not show up.

jfboismenu-adsk:
Reimplemented #28769 since it broke on pre-6.0 sites. We're now validating that the field exists before trying to filter on it.

### DIFF
--- a/python/tk_desktop/project_model.py
+++ b/python/tk_desktop/project_model.py
@@ -217,19 +217,30 @@ class SgProjectModel(ShotgunModel):
 
     _supports_project_templates = None
 
-    @staticmethod
-    def supports_project_templates(connection):
+    @classmethod
+    def supports_project_templates(cls, connection):
+        """
+        Tests if Shotgun 6.0 Project Templates are supported on the server. If
+        this method has never been called, the server will be contacted
+        synchronously and the result will be cached so subsequent calls are
+        faster.
+
+        :param connection: A Shotgun connection instance.
+
+        :returns: True if the server supports Shotgun 6.0 Project Templates,
+                  False otherwise.
+        """
         # If we haven't checked on the server yet.
-        if SgProjectModel._supports_project_templates is None:
+        if cls._supports_project_templates is None:
             try:
                 # Try to read the field's schema.
                 connection.schema_field_read("Project", "is_template")
-                # If worked therefore it exists.
-                SgProjectModel._supports_project_templates = True
+                # It worked therefore it exists.
+                cls._supports_project_templates = True
             except shotgun_api3.Fault:
                 # We got a fault, so it doesn't exist.
-                SgProjectModel._supports_project_templates = False
-        return SgProjectModel._supports_project_templates
+                cls._supports_project_templates = False
+        return cls._supports_project_templates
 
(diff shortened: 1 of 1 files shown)


python/tk_desktop/project_model.py
```diff
@@ -19,6 +19,7 @@
 
 shotgun_model = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_model")
 shotgun_login = sgtk.platform.import_framework("tk-framework-login", "shotgun_login")
+shotgun_api3 = sgtk.platform.import_framework("tk-framework-login", "shotgun_api3")
```
robblau: This seems like it should be

```python
from tank_vendor import shotgun_api3
```

Simpler and we're getting out of the business of the login framework handling connections, which this feels related to.

python/tk_desktop/project_model.py
```diff
@@ -214,6 +215,22 @@ class SgProjectModel(ShotgunModel):
     thumbnail_updated = QtCore.Signal(QtGui.QStandardItem)
     project_launched = QtCore.Signal()
 
+    _supports_project_templates = None
+
+    @staticmethod
```
robblau: Not sure if the difference is intentional (to lock subclasses from being able to change this behavior) but I'd expect this to be

```python
@classmethod
def supports_project_templates(cls, connection):
...
```

And then replace the ```SgProjectModel``` explicit symbols in the definition with ```cls```.
Is there a reason to make it static as opposed to class?

python/tk_desktop/project_model.py
```diff
@@ -214,6 +215,22 @@ class SgProjectModel(ShotgunModel):
...
+    def supports_project_templates(connection):
+        # If we haven't checked on the server yet.
```
robblau: Doc string?  We should include the reason why we are doing this check (Template projects in 6.0) so whoever looks at this 5 years from now understands the motivation for the check.

python/tk_desktop/project_model.py
```diff
@@ -232,6 +249,11 @@ def __init__(self, parent, overlay_parent_widget):
             ["archived", "is_not", True],
             ["is_template", "is_not", True]
```
robblau: ```is_template``` should be removed from the default list, no?

python/tk_desktop/project_model.py
jfboismenu-adsk: That's what I did at first. Unfortunately, the exception being thrown is from tk-framework-login's version of shotgun_api3. Therefore, if I catch the Fault exception from tank_vendor.shotgun_api3, it passes right through... That's why I am stuck catching the one from tk-framework-login.shotgun_api3.

python/tk_desktop/project_model.py
jfboismenu-adsk: Sorry, old habits die hard. I only started using @classmethod recently and therefore often forget about it. The code looks a lot better using @classmethod now. Thanks.

python/tk_desktop/project_model.py
jfboismenu-adsk: Forgot to commit my file before pushing. :(

PR #17, opened 2015-04-10
#27965 - Improved Toolkit Security

manneohrstrom:
Creating a pull request for all the security changes, mainly for CR purposes.

### DIFF
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -778,7 +778,12 @@ class DesktopWindow(SystrayWindow):
             "core_python_path": core_python,
             "config_path": config_path,
             "project": project,
-            "current_user": sg_auth.serialize_user(sgtk.get_authenticated_user()),
+            "current_user": sg_auth.serialize_user(
+                # We'll be launching the background processes as the highest privilege user
+                # available. CoreDefaultsManager will make sure a script user is used to launch the
+                # DCCs if it is available.
+                sg_auth.ShotgunAuthenticator(sgtk.util.CoreDefaultsManager()).get_default_user()
+            ),
             "proxy_data": {
                 "proxy_pipe": server_pipe,
                 "proxy_auth": server_auth,
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -410,8 +410,6 @@ class DesktopWindow(SystrayWindow):
         engine = sgtk.platform.current_engine()
 
         try:
-            # FIXME: Need to clear the password in the keychain here when that
-            # functionality is brought back.
             if engine.uses_legacy_authentication():
                 engine.create_legacy_login_instance().logout()
             else:
@@ -778,12 +776,6 @@ class DesktopWindow(SystrayWindow):
             "core_python_path": core_python,
             "config_path": config_path,
             "project": project,
-            "current_user": sg_auth.serialize_user(
-                # We'll be launching the background processes as the highest privilege user
-                # available. CoreDefaultsManager will make sure a script user is used to launch the
-                # DCCs if it is available.
-                sg_auth.ShotgunAuthenticator(sgtk.util.CoreDefaultsManager()).get_default_user()
-            ),
             "proxy_data": {
                 "proxy_pipe": server_pipe,
                 "proxy_auth": server_auth,
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -412,7 +412,10 @@ class DesktopWindow(SystrayWindow):
         try:
             # FIXME: Need to clear the password in the keychain here when that
             # functionality is brought back.
-            sg_auth.ShotgunAuthenticator().clear_default_user()
+            if engine.uses_legacy_authentication():
+                engine.create_legacy_login_instance().logout()
+            else:
+                sg_auth.ShotgunAuthenticator().clear_default_user()
         except Exception:
             # if logout raises an exception, just log and don't crash
             engine.log_exception("Error logging out")
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -93,7 +93,7 @@ class DesktopWindow(SystrayWindow):
         self.ui.apps_button.style().unpolish(self.ui.apps_button)
         self.ui.apps_button.style().polish(self.ui.apps_button)
 
-        interactive_authentication.ui_authenticate(force_human_user_authentication=True)
+        interactive_authentication.ui_authenticate()
         connection = shotgun.get_sg_connection()
 
         engine = sgtk.platform.current_engine()
@@ -133,7 +133,9 @@ class DesktopWindow(SystrayWindow):
         self.user_menu.addAction(self.ui.actionRefresh_Projects)
         about_action = self.user_menu.addAction("About...")
         self.user_menu.addSeparator()
-        self.user_menu.addAction(self.ui.actionSign_Out)
+        # Only show sign out if you are logged as a human user.
+        if authentication.is_human_user_authenticated():
+            self.user_menu.addAction(self.ui.actionSign_Out)
         self.user_menu.addAction(self.ui.actionQuit)
 
         name_action.triggered.connect(self.open_site_in_browser)
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -20,7 +20,10 @@ from tank.platform.qt import QtCore, QtGui
 
 import sgtk
 from sgtk.util import shotgun
+from sgtk.util import login
+from sgtk.util import authentication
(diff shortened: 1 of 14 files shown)


python/tk_desktop/desktop_window.py
```diff
@@ -20,7 +20,9 @@
 
 import sgtk
 from sgtk.util import shotgun
+from sgtk.util import login
 from sgtk.platform import constants
+from tank_vendor import shotgun_authentication as sg_auth
```
manneohrstrom: this seems to be used only in one place in the code. I suggest using the full name instead!

python/tk_desktop/desktop_window.py
```diff
@@ -40,6 +42,7 @@
 from .project_commands_model import ProjectCommandModel
 from .project_commands_model import ProjectCommandProxyModel
 from .project_commands_widget import ProjectCommandDelegate
+from shotgun_desktop.desktop_login import get_shotgun_authenticator
```
manneohrstrom: So this imports code from inside of the desktop executable python repo. I looked at that code and it is not really obvious that this method is part of a public API that can be accessed by external things. Does the desktop-internal API have a "public API" at this point already that we can add `get_shotgun_authenticator` to? I am worried that things will break because that method is changed in desktop-internal and it is not obvious that it is called from somewhere else! I think it could be useful to clarify this - maybe by establishing a "public API" in desktop internal. What do you think?

python/tk_desktop/desktop_window.py
```diff
@@ -92,8 +93,8 @@ def __init__(self, parent=None):
         self.ui.apps_button.setProperty("active", True)
         self.ui.apps_button.style().unpolish(self.ui.apps_button)
         self.ui.apps_button.style().polish(self.ui.apps_button)
-        login = ShotgunLogin.get_instance_for_namespace("tk-desktop")
-        connection = login.get_connection()
+
+        connection = shotgun.get_sg_connection()
```
manneohrstrom: ideally, it would be nice if this could be `sgtk.util.get_shotgun_connection()` and if we documented it in the offcial API docs (here: https://toolkit.shotgunsoftware.com/entries/23808858#Utility%20Functions ).

python/tk_desktop/desktop_window.py
```diff
@@ -135,7 +133,9 @@ def __init__(self, parent=None):
         self.user_menu.addAction(self.ui.actionRefresh_Projects)
         about_action = self.user_menu.addAction("About...")
         self.user_menu.addSeparator()
-        self.user_menu.addAction(self.ui.actionSign_Out)
+        # Only show sign out if you are logged as a human user.
+        if sg_auth.is_session_user(sgtk.get_current_user()):
```
manneohrstrom: Is this possible? Could you be logged in as something else?
I also thought we decided to try to just have a user object and not distinguish between a script user and a human user outside of the security API?

python/tk_desktop/desktop_window.py
```diff
@@ -505,9 +507,11 @@ def show_update_project_config(self):
         self.update_project_config_widget.show()
         self.project_overlay.hide()
 
+    def _get_current_user(self):
+        return login.get_current_user(sgtk.platform.current_engine().tank)
```
manneohrstrom: This should ideally use the official, public, API method `sgtk.util.get_current_user()` - see the docs here - https://toolkit.shotgunsoftware.com/entries/23808858#Utility%20Functions

python/tk_desktop/desktop_window.py
```diff
@@ -843,9 +847,7 @@ def slide_finished():
         anim_group.start()
 
     def open_site_in_browser(self):
-        login = ShotgunLogin.get_instance_for_namespace("tk-desktop")
-        connection = login.get_connection()
-        url = connection.base_url
+        url = shotgun.get_sg_connection().base_url
```
manneohrstrom: does this mean that there needs to be a connection in order to get the url? Didn't we introduce a convienence method that allowed you to just pull this from the config file directly to avoid an overhead? Or maybe that doesn't matter in this case?

python/tk_desktop/project_commands_model.py
```diff
@@ -267,7 +264,7 @@ def _handle_command_triggered(self, item, command_name=None, button_name=None,
             "event_type": self.APP_LAUNCH_EVENT_TYPE,
             "project": self.__project,
             "meta": {"name": command_name, "group": group_name},
-            "user": login,
+            "user": login.get_current_user(engine.sgtk),
```
manneohrstrom: see comment earlier!

python/tk_desktop/desktop_window.py
manneohrstrom: it also looks a little bit funny when you mix sgtk and tank :)

python/tk_desktop/project_commands_model.py
```diff
@@ -318,12 +315,9 @@ def __initialize_recents(self):
         # and a boolean saying whether the corresponding command has been registered
         self.__recents = {}
 
-        # need to know what login to find events for
-        login = ShotgunLogin.get_instance_for_namespace("tk-desktop").get_login()
-
         # pull down matching invents for the current project for the current user
         filters = [
-            ["user", "is", login],
+            ["user", "is", login.get_current_user(sgtk.platform.current_engine().tank)],
```
manneohrstrom: please use sgtk and not tank

manneohrstrom: what if `shotgun.yml` contains script credentials? Does that mean that it will use those, and the login core hook will then in turn be used? And does that mean that this may be a regression compared with the previous behaviour?

The previous code pulls the login out of the login framework? Should the new code follow the same pattern and pull the login info out of the shotgun_authentication module rather than sgtk?

python/tk_desktop/project_model.py
```diff
@@ -245,8 +244,7 @@ def __init__(self, parent, overlay_parent_widget):
         """ Constructor """
         ShotgunModel.__init__(self, parent, download_thumbs=True)
 
-        # override the default connection to use one that is tied to the current login
-        connection = ShotgunLogin.get_instance_for_namespace("tk-desktop").get_connection()
+        connection = sgtk.platform.current_engine().sgtk.shotgun
```
manneohrstrom: The engine has a `shotgun` property that you can use! - e.g. `sgtk.platform.current_engine().shotgun`

python/tk_desktop/desktop_window.py
robblau: Maybe this waits for the upgradable bootstrap logic for desktop internal, but shouldn't this be pulling from something like the login framework so that we can centralize the code that does the caching of these values?  That way in the future as we have better support for multiple logins and encrypted storage it will be easier to update?

Seems like ideally no authentication logic would be in tk-desktop or tk-desktop-internal.  Is there a way we can implement this to get closer to that?

python/tk_desktop/project_model.py
```diff
@@ -286,17 +284,16 @@ def _before_data_processing(self, sg_data_list):
         # merge in timestamps from project launch events if they are newer than
         # the last access timestamps from Shotgun
 
+        engine = sgtk.platform.current_engine()
         # pull down matching events for the current user
-        login = ShotgunLogin.get_instance_for_namespace("tk-desktop").get_login()
         filters = [
-            ["user", "is", login],
+            ["user", "is", login.get_current_user(engine.tank)],
```
manneohrstrom: ideally use `engine.sgtk` here

python/tk_desktop/desktop_window.py
manneohrstrom: Good point! 
The reason there is a dependency is that desktop-internal has its own defaults manager which handles some of the desktop specific logic for how default values etc are saved. This is different in the desktop vs core - the desktop using the login fw/keychain vs core picking up some stuff from shotgun.yml. Eventually, if desktop and core converges on this functionality, the external module will take care of everything. Right now, since there are differences, core and desktop-internal each implements a small class to define certain behaviour around storing of defaults etc. 

But the code executed right here is just handling logging out. I wonder if the code could simply be

```
from tank_vendor.shotgun_authentication import ShotgunAuthenticator
shotgun_auth = ShotgunAuthenticator()
shotgun_auth.clear_saved_user()
```

manneohrstrom:
CR done! Had some general questions!

python/tk_desktop/desktop_window.py
jfboismenu-adsk: Yeah, that's a better approach than the one I am taking. I was simply translating the code from the old way of doing things to the new one. But yours is better. And the shotgun.get_associated_sg_base_url() is not new functionality, it's been there for a while: get_associated_sg_config_data is.

python/tk_desktop/desktop_window.py
jfboismenu-adsk: I didn't notice that tank was aliased to sgtk in bundle.py. Thanks!

PR #18, opened 2015-04-17
#29165 If interpreter is not found, the "set up project for a new OS" is displayed

manneohrstrom:
This change will display the "setup project for new operating system" whenever the python interpreter file is not found. Previously, it would only show if no value had been defined in the interpreter was found (which corresponded to an uninitialized OS). We are now seeding all OSes with default interpreter paths to help make things work out of the box, meaning that the case where there is a blank interpreter file is much less common. Also added more logging around this to help support and debugging.

This change could be confusing if a user has installed desktop in one location and the rest of her/his studio has installed it somewhere else. This user would now be presented with the "set up for a new OS" screen rather than the (much more technical looking) error message that was previously shown.

I personally think this is ok, but I guess we could polish the copy slightly on the custom UI splash page to more accurately capture this case. Curious to hear what @robblau thinks!

### DIFF
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -732,7 +732,7 @@ class DesktopWindow(SystrayWindow):
                         # python not specified for this os, show the setup new os widget
                         engine.log_error("Cannot find interpreter '%s' defined in "
                                          "config file %s. Will show the special "
-                                         "'setup new OS' screen" % (path_to_python, interpreter_config_file))
+                                         "'no python' UI screen" % (path_to_python, interpreter_config_file))
                         self.setup_new_os_widget.show()
                         self.project_overlay.hide()
                         return
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -728,15 +728,14 @@ class DesktopWindow(SystrayWindow):
                         path_to_python = f.read().strip()
                         core_root = current_config_path
 
-                    if not path_to_python:
+                    if not path_to_python or not os.path.exists(path_to_python):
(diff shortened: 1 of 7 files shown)


robblau:
I do like the intent of the change.  I do think that given the change we need to give a little more context to the user about what went wrong... since now there is a possible setup/install issue that'll be masquerading as an unsetup operating system.

Do you think there is a better check now for "this isn't setup for this OS" (checking the storage for whether the path for the OS is specified)?

At the very least, lets add some details to the setup_new_os widget so that if python path is defined and doesn't exist.  Smaller text that blends into the background a bit more that says ```Python not found at "C:\You\Should\Have\Installed\Here".```

manneohrstrom:
perfect, i'll update that UI. That feels like the right way to me too. I think we also have a ticket about the fonts being too large on windows for that screen, i'll see if we can tweak that too!

manneohrstrom:
UI updated to show the following:

screen shot 2015-04-24 at 14 15 41

Please note that this error message is now much less common - previously, all users who tried to run a project on two platforms would see the message. Now, it will only be displayed for people who either have some advanced setup or who have desktop installed in a non-standard location.

Does this look ok? If so then I will go ahead and update the docs too.

manneohrstrom:
Brief note:
- also fixed the copyright notice saying 2014
- also fixed the inconsisntencies in font size between Operating systems (#26571)

robblau:
Looks much better!  I think that will definitely help anybody who runs into the issue have a better idea of what is going on.

manneohrstrom:
Awesome! I'll proceed with merging this in and updating the documentation! cheers!

PR #19, opened 2015-05-01
For #29058, Maintain backward compatibility with old Desktop app

jfboismenu-adsk:
- Added code that reads the old credentials from the tk-desktop-internal bootstrap code and
  injects them into the new one.

### DIFF
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -11,6 +11,8 @@
 import os
 import re
 import sys
+import time
+import subprocess
 import string
 import logging
 import collections
@@ -18,7 +20,8 @@ import collections
 from sgtk.errors import TankEngineInitError
 
 from . import rpc
-from . import desktop_window
+from distutils.version import StrictVersion
+import sgtk
 
 
 class DesktopEngineSiteImplementation(object):
@@ -188,6 +191,10 @@ class DesktopEngineSiteImplementation(object):
         """
         self.app_version = version
 
+        if self.uses_legacy_authentication():
+            self._assert_authentication_fault_exists(splash)
+            self._migrate_credentials(splash)
+
         # Initialize Qt app
         from tank.platform.qt import QtGui
 
@@ -217,6 +224,17 @@ class DesktopEngineSiteImplementation(object):
         f.close()
         app.setStyleSheet(css)
 
+        # desktop_window needs to import shotgun_authentication globally. However, doing so
+        # can cause a crash when running pre-1.0.2 bootstrap code. We used to not restart
+        # Desktop when upgrading the core, which caused the older version of core to be
+        # kept in memory and the newer core to not be used until the app was reloaded.
+        # Since pre 0.16 cores didn't have a shotgun_authentication module, we
+        # would have crashed if this had been imported at init time. Note that earlier
+        # in this method we forcefully restarted the application if we noticed
+        # that the core was upgraded without restarting. Which means that if we
+        # end up here, it's now because we're in a good state.
+        from . import desktop_window
+
         # initialize System Tray
         self.desktop_window = desktop_window.DesktopWindow()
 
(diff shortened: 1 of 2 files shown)


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -231,6 +249,79 @@ def run(self, splash=None, version=None):
...
         return result
 
+    def uses_legacy_authentication(self):
+        """
+        Returns if the Desktop's bootrap code uses the tk-framework-login for
+        logging in.
+
+        :returns: True the bootstrap logic is older than 1.1.0, False otherwise.
+        """
+        return StrictVersion(self.app_version) < StrictVersion("1.1.0")
```
manneohrstrom: In other places of the code, we are using LooseVersion, just to cater for version numbers with RC suffixes, etc. Also, would it make sense to use the util methods here? https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/deploy/util.py#L115

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -231,6 +249,79 @@ def run(self, splash=None, version=None):
...
+        """
+        try:
+            from python import ShotgunLogin
+        except ImportError:
+            self.log_exception("Could not import tk-framework-login")
+            raise
+        else:
+            return ShotgunLogin.get_instance_for_namespace("tk-desktop")
+
+    def _assert_authentication_fault_exists(self, splash):
```
manneohrstrom: I think the word "restart" should be part of this method somehow. Does that make sense? Assert feels more like a check that is commented out in a release build - whereas this method is pretty drastic in comparison - it might restart the entire application! :) 

Although we are probing some stuff inside the Shotgun API, am I right thinking that it is the site config core API that is too old and this is why we are restarting? Maybe it should be called `_force_restart_on_old_core()` or something?

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -231,6 +249,79 @@ def run(self, splash=None, version=None):
...
+        """
+        Asserts that we importing the right version of shotgun_api3. If we are
+        not, the user will be shown a countdown and the app will be restarted.
+        """
+        from tank_vendor import shotgun_api3
```
manneohrstrom: Is there any case you can think of where the user ends up in an infinite loop? What if the internet connection is down and the core auto-update raises an error? Is this a case we should QA?

If I am reading this right, and we are trying to establish if we have a more recent core API available because it was downloaded (but not loaded in memory yet), maybe the logic could be something like this:

```
if the core in memory is pre 0.16:
   if 0.16 is available on disk:
       countdown splash + restart_desktop
   else:
       display error message
```

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -231,6 +249,79 @@ def run(self, splash=None, version=None):
         result = app.exec_()
         return result
 
+    def uses_legacy_authentication(self):
+        """
+        Returns if the Desktop's bootrap code uses the tk-framework-login for
```
manneohrstrom: IMHO "Desktop" in this context is confusing because it can refer to both the tk-desktop engine and the installed desktop. Maybe be explicit and say "Shotgun Desktop's installed code" or something?

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -231,6 +249,79 @@ def run(self, splash=None, version=None):
...
+        not, the user will be shown a countdown and the app will be restarted.
+        """
+        from tank_vendor import shotgun_api3
+        try:
+            # Try to access the new AuthenticationFault exception. If we are
+            # running a pre-1.0.2 version of Desktop, Desktop will not have been
+            # rebooted after the core upgrade and we'll get the pre 0.16.0
+            # shotgun_api3 module, which doesn't have the AuthenticationFault
+            # class.
+            shotgun_api3.AuthenticationFault
```
manneohrstrom: Should we be worried about this check? Are we essentially using this to determine if the a core distribution is v0.16 or earlier? Would it make sense to for example read the `info.yml` file that is at the top level of the core package? This is what we normally use to check the version of the core.

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -231,6 +249,79 @@ def run(self, splash=None, version=None):
...
+        Returns if the Desktop's bootrap code uses the tk-framework-login for
+        logging in.
+
+        :returns: True the bootstrap logic is older than 1.1.0, False otherwise.
+        """
+        return StrictVersion(self.app_version) < StrictVersion("1.1.0")
+
+    def create_legacy_login_instance(self):
+        """
+        Creates the pre-bootstrap-1.1.0 ShotgunLogin used to authenticate.
```
manneohrstrom: Could this be a little more descriptive? Just to help future readers of this code. When we re-read this code 8 months from now and have long forgotten about the security work, a slightly more verbose comment would probably be useful just to summarize. The current docstring explains the "what" - which is great. But with compatibility stuff like this, also providing a short "why" often helps when going back to the code. Hope it makes sense!

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -231,6 +249,79 @@ def run(self, splash=None, version=None):
...
+
+        :raises ImportError: Thrown if the module is not available.
+
+        :returns: A tk-framework-login's ShotgunLogin instance.
+        """
+        try:
+            from python import ShotgunLogin
```
manneohrstrom: a comment here saying something like "this only works on v1.x of the desktop executable, in desktop v2.x it has been removed etc etc etc" could be helpful I think. I guess I am missing a small explanation *why* we are expecting the import to raise an error.

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -231,6 +249,79 @@ def run(self, splash=None, version=None):
...
+            # class.
+            shotgun_api3.AuthenticationFault
+        except AttributeError:
+            # Provide a countdown so the user knows that the desktop app is
+            # being restarted on purpose because of a core update. Otherwise,
+            # the user would get a flickering splash screen that from the user
+            # point of view looks like the app is redoing work it already did by
+            # mistake. This makes the behavior explicit.
+            for i in range(3, 0, -1):
+                splash.set_message("Core updated. Restarting desktop in %d seconds..." % i)
```
manneohrstrom: i think splash can be `None`.

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -231,6 +249,79 @@ def run(self, splash=None, version=None):
...
+        except ImportError:
+            self.log_exception("Could not import tk-framework-login")
+            raise
+        else:
+            return ShotgunLogin.get_instance_for_namespace("tk-desktop")
+
+    def _assert_authentication_fault_exists(self, splash):
+        """
+        Asserts that we importing the right version of shotgun_api3. If we are
+        not, the user will be shown a countdown and the app will be restarted.
```
manneohrstrom: Missing splash param.

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -231,6 +249,79 @@ def run(self, splash=None, version=None):
...
+            for i in range(3, 0, -1):
+                splash.set_message("Core updated. Restarting desktop in %d seconds..." % i)
+                time.sleep(1)
+            subprocess.Popen(sys.argv)
+            sys.exit(0)
+
+    def _migrate_credentials(self, splash):
+        """
+        Migrates the credentials from tk-framework-login to
+        shotgun_authentication.
```
manneohrstrom: Missing splash param.

manneohrstrom: Actually do we need splash? Looks like it may not be used?

manneohrstrom:
CR done. Some comments -- i think it's mostly semantics and wording. This change handles some pretty intricate logic across different version of various components. I think we can help code readers by providing a bit more explanation around these cases, explaining why we are doing some stuff. Just to help future code readers (most likely ourselves!) understand the flow around some of this.

python/tk_desktop/desktop_engine_site_implementation.py
jfboismenu-adsk: The else branch is not possible. The tk-desktop 2.0.0 engine will download if and only if you have Core 0.16.0 and higher. So basically, the test is good enough I think and less complicated than finding an info.yml file, parsing it and comparing values.

If the Internet is down, then you won't download the core 0.16 and you won't be running the tk-desktop 2.0

python/tk_desktop/desktop_engine_site_implementation.py
jfboismenu-adsk: The method seems to advertise that fact, but in reality it is always set. I should update the signature to not accept None.

python/tk_desktop/desktop_engine_site_implementation.py
jfboismenu-adsk: Should we? Yes. Yes. :)

This felt more straightforward than going on disk, reading up a file and getting the version number from it and then compare. It will, however, break if we ever remove AuthenticationFault for some reason (which I bet is highly unlikely since once the code is out, we can't easily remove it).

python/tk_desktop/desktop_engine_site_implementation.py
manneohrstrom: cool. It seemed a little odd to me why there was a `def run(self, splash=None, version)`!

PR #21, opened 2015-06-15
Assigns user to the template project.

jfboismenu-adsk:


python/tk_desktop/assign_users_to_template_project_command.py
```diff
@@ -0,0 +1,51 @@
...
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sgtk
+
+
+name = "Assign users to Template Project"
+properties = {
+    "short_name": "assign_users_to_template_project",
+    "description": "Assign all users to the Template Project. This allows the user to "
```
manneohrstrom: looks like part of the description is missing!

python/tk_desktop/assign_users_to_template_project_command.py
```diff
@@ -0,0 +1,51 @@
...
+        return False
+    return True
+
+
+def callback():
+    """
+    Assigns all human users to the pipeline configuration's project.
+    """
+
+    engine = sgtk.platform.current_engine()
```
manneohrstrom: I think it would be good to have a confirm dialog here explaining what will happen (and why) just to keep the user in the loop!

python/tk_desktop/assign_users_to_template_project_command.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # Retrieve all the users.
+    users = engine.shotgun.find("HumanUser", [])
+
+    # Assign them to the current Project.
+    engine.shotgun.update(
+        "Project",
+        pc.get_project_id(),
+        {"users": users}
+    )
```
manneohrstrom: What sort of feedback will you be getting when you are running this command?
I think it could make sense to pop up another dialog to confirm that the command has completed.
Similarly, what if it fails? I think that too should be trapped in a dialog?

python/tk_desktop/assign_users_to_template_project_command.py
```diff
@@ -0,0 +1,51 @@
...
+def is_available(engine):
+    """
+    Returns if this command should be available.
+
+    :param engine: The current engine.
+
+    :returns: True if the pipeline configuration stll has a Project assigned to it, False otherwise.
+    """
+    # If there is no project assigned, there's nothing to do.
+    if engine.sgtk.pipeline_configuration.get_project_id() is None:
```
manneohrstrom: Or `if engine.sgtk.pipeline_configuration.is_site_configuration()`  ?

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -226,9 +232,24 @@ def run(self, splash, version, **kwargs):
...
 
+        # We've retrieved the permission rule set for immediate consumption, so
+        # copy the value and remove it from the user dictionary.
+        permission_rule_set = self._current_login[_PERMISSION_RULE_SET]
+        del self._current_login[_PERMISSION_RULE_SET]
+
+        # If the permission rule set is for the admin, we'll register a few
+        # commands.
+        if permission_rule_set["name"] == _ADMIN_RULE_SET:
+            if assign_users_to_template_project_command.is_available(self._engine):
```
manneohrstrom: I am not super familiar with this code, but the class name says `DesktopEngineSiteImplementation`, implying that this is site specific. So do we need the `is_available()` check? Will it ever return false? If it does return false in some cases, a comment explaining could be helpful I think!

python/tk_desktop/desktop_window.py
```diff
@@ -125,10 +125,24 @@ def __init__(self, parent=None):
...
+        # Build a menu with all the commands
+        commands = engine.commands
+        if commands:
+            self.user_menu.addSeparator()
+            for name, command in commands.iteritems():
+                props = command["properties"]
+                action = QtGui.QAction(self)
+                action.setObjectName(props["short_name"])
+                action.setText(name)
+                action.setToolTip(props.get("description", "No description available."))
```
manneohrstrom: I think it's probably better to just omit the tooltip rather than having it explicit say "no info available".

python/tk_desktop/assign_users_to_template_project_command.py
```diff
@@ -0,0 +1,51 @@
...
+def callback():
+    """
+    Assigns all human users to the pipeline configuration's project.
+    """
+
+    engine = sgtk.platform.current_engine()
+    pc = engine.sgtk.pipeline_configuration
+
+    # Retrieve all the users.
+    users = engine.shotgun.find("HumanUser", [])
```
manneohrstrom: I think we should filter out all non-active users.

manneohrstrom:
CR done. Some questions:

- I think

manneohrstrom:
CR done. Some questions and thoughts:

- We are asking the admin to do some obscure stuff here because we essentially broke stuff in one of our updates. The whole approach here is a little unusual, so needs to be backed up by clear instructions what is going on I think, and a followup once it has completed.

- What's the story around sites that don't have the template project associated? I am guessing the button shows for those as well right now? I think we should hide it if possible.

Thanks!

python/tk_desktop/assign_users_to_template_project_command.py
jfboismenu-adsk: No. I want this to be used by as many people as possible. Therefore I can't put a dependency on features that aren't released yet.

manneohrstrom: But this change to tk-desktop will be downloaded through the auto update of desktop right?
So by that time, people would have downloaded core?
At the very least, please put a comment in explaining this. It's not really obvious.

jfboismenu-adsk: That core hasn't been released yet actually! :) This is meant to be released ASAP. I'll add the comment tough.

jfboismenu-adsk:
The menu item doesn't display for a user who has a pipeline configuration with the None id, since it means the project has been migrated by the admin. This is what the is_available() check is for.

PR #22, opened 2015-06-19
For #28904, two factor authentication support in old desktop

jfboismenu-adsk:
- Adds support for two factor authentication when using the old installer with the new desktop engine.

### DIFF
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -324,24 +324,36 @@ class DesktopEngineSiteImplementation(object):
         shotgun_authentication.
         """
         sl = self.create_legacy_login_instance()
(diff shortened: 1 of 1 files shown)


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -324,24 +324,37 @@ def _migrate_credentials(self):
...
-            # Ugly, but this is the only way available to get at the
-            # raw http_proxy string.
-            http_proxy=sl._http_proxy
-        )
 
         # Next set the current host and user in the framework.
         dm = DefaultsManager()
-        dm.set_host(user.host)
-        dm.set_login(user.login)
+        dm.set_host(site)
```
jfboismenu-adsk: This was actually a bug. If we don't set the current host first, create_session_user will then cache the user inside the wrong site.

PR #23, opened 2015-09-16
For #31081, cleanly shutdown websockets server

jfboismenu-adsk:
- Launching subprocesses with the close_fds flag to prevent inheriting socket handles, which meant that subprocesses could hold the file descriptor for the port 9000 open.
- Adding desktop server logger to the desktop log


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -206,6 +206,10 @@ def run(self, splash, version, **kwargs):
         # this engine.
         self.startup_version = kwargs.get("startup_version")
 
+        server = kwargs.get("server")
```
robblau: Comments about what is going on here would be good.

PR #24, opened 2015-10-01
#31384 Add Tabs for App Registered Panels

jesseemond-adsk:
When an application used by `tk-desktop` registers a panel, this panel is now initialized and its widget is added as a tab in the desktop application view. A button is also added to allow switching to that tab.

Here is the result with a `tk-multi-starterapp` modified to register itself as a panel (with title `My Tasks`):

screen shot 2015-10-01 at 3 31 39 pm

For the moment, the `Apps` tab is manually added in the engine, although it might be interesting to consider extracting it as an app later on.

A couple of UI details:
- removed the `Apps` button, as it is now automatically generated like any other tab button.
- renamed the `stack` `QStackWidget` to `apps_tab`, to clarify its purpose.
- added a `QStackWidget` called `tab_view` as a parent to `apps_tab` to hold the tab widgets and only show the current tab.
- my version of PySide is `1.2.2`, so it updated the generated ui files. Hope this is not a problem.
- modification of the *css* was required to give a distinct look to active/unactive tabs. Also took on the opportunity to style hover/pressed.


jesseemond-adsk:
Moved to https://github.com/shotgunsoftware/tk-desktop/pull/25 to merge on a separate branch until the functionality is confirmed to work with the `My Tasks` app.

PR #25, opened 2015-10-01
#31384 Add Tabs for App Registered Panels

jesseemond-adsk:
When an application used by `tk-desktop` registers a panel, this panel is now initialized and its widget is added as a tab in the desktop application view. A button is also added to allow switching to that tab.

Here is the result with a `tk-multi-starterapp` modified to register itself as a panel (with title `My Tasks`):

screen shot 2015-10-01 at 3 31 39 pm

For the moment, the `Apps` tab is manually added in the engine, although it might be interesting to consider extracting it as an app later on.

A couple of UI details:
- removed the `Apps` button, as it is now automatically generated like any other tab button.
- renamed the `stack` `QStackWidget` to `apps_tab`, to clarify its purpose.
- added a `QStackWidget` called `tab_view` as a parent to `apps_tab` to hold the tab widgets and only show the current tab.
- my version of PySide is `1.2.2`, so it updated the generated ui files. Hope this is not a problem.
- modification of the *css* was required to give a distinct look to active/unactive tabs. Also took on the opportunity to style hover/pressed.


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -48,6 +48,35 @@ def destroy_engine(self):
...
             self.msg_server.close()
 
+
+    ###########################################################################
+    # panel support (displayed as tabs)
+    def _init_tabs(self):
+        """ Create desktop tabs from the registered application panels """
+        for panel in self._engine.panels.itervalues():
+            # let the panel show itself through its registered callback
+            panel['callback']()
```
robblau: Should use double quotes to match coding conventions.

python/tk_desktop/desktop_window.py
```diff
@@ -284,6 +282,59 @@ def _pop_dll_state(self):
                 engine = sgtk.platform.current_engine()
                 engine.log_warning('Could not restore DllDirectory under Windows.')
 
+    def register_tab(self, tab_name, tab_widget):
+        '''
```
robblau: Double quotes again.

python/tk_desktop/desktop_window.py
```diff
@@ -284,6 +282,59 @@ def _pop_dll_state(self):
...
+        # tab-specific values
+        tab_button.setText(tab_name)
+
+        # setup the tab widget
+        tab_widget.setParent(self)
+
+
+        # define the event handler when the user changes tab
+        def on_tab_selected():
+            '''
```
robblau: Double quotes.

python/tk_desktop/desktop_window.py
jesseemond-adsk: @robblau I used the file convention in this case. Should I change the whole file at the same time?

robblau:
Panels with Desktop are a little different from other apps, since with Desktop there is an ordering to the panels.  It would be good to add in an optional argument to the panel registration that lets you ask for a particular spot in the app.  This would let the apps be ordered.  It should be possible to put something before "Apps" even.

jesseemond-adsk:
@robblau how about a number representing a relative position?

- `Apps` could be 0 for now, until it becomes an app on its own. 0 would make it first, unless another app has a negative position.
- panels registering without a specified position could be given a number by default (I'm thinking their position in the *panels* list to place them at the end by default)
- panels would be added as tabs in the order specified by their position

A problem I would see with that approach is customizing tab ordering: would that imply forking the apps to change the passed position? A simple solution I would see for that is adding a config option to specify an index (`tab_position`?) to apps that we intend to use in a tab, and pass that config value to `register_panel`.

Would you have a better solution?

robblau:
I think it is fine to start with something that would require a fork to change the default ordering of a panel.  If we use floats rather than ints then any app devs could put a panel wherever they want, they just couldn't change the relative ordering of our apps without forking them... which is fine unless we get a request that people really want the ordering to be different.

Agreed that no specified ordering should just go to the end of the list.

If we ever need to add a setting for tab ordering, feels like it should be on desktop rather than on the app itself so that the logic for all of this keeps existing in one place.

jesseemond-adsk:
Here is an example of a sample app setting its `position` as -1:
screen shot 2015-10-02 at 9 23 11 am

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -48,6 +49,78 @@ def destroy_engine(self):
...
+        """
+        Create desktop tabs from the registered application panels, following
+        their specified relative positioning.
+        """
+        # make the "Apps" tab appear as a fake panel to fit in the sorting
+        # logic
+        apps_tab = {"callback":   self.desktop_window._register_apps_tab,
+                    "properties": {"position": 0}}
+
+        panels = [apps_tab] + list(self._engine.panels.itervalues())
```
robblau: Rather than calling list(dict().itervalues()) you may as well just call dict().values(), since you're keeping the whole list in memory at that point anyway.

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -48,6 +49,78 @@ def destroy_engine(self):
...
+
+        def panel_position(panel):
+            """ Returns the relative position of a panel """
+            # sane checking on the expected structure
+            assert("properties" in panel, "panel should have properties set")
+            assert("position" in panel["properties"],
+                   "panel should have a position set")
+            return panel["properties"]["position"]
+
+        ordered_panels = sorted(panels, key=panel_position)
```
robblau: I don't think this logic gets the outcome we talked about.  We are going for any apps with position set will come first in position order, and then the rest will show up in whatever order the register gets called, right?

If something is registered with position 1000, it would probably come last with this logic.

Seems like keeping position optional in the dictionary and defining a cmp that just orders according to the above criteria would give the right ordering.

jesseemond-adsk: @robblau Oh, I misunderstood the intended behaviour. Will update.
As for the "automatic" ordering of unspecified positions, how about an `auto_position` property with the automatic numbering? Otherwise we'll get surprising default ordering since the panels are stored in a dictionary with their names as keys (i.e. they're not in the same order as the registers happened).
(We could also store panels in an `OrderedDict` instead, but that would be in `tk-core`)

robblau: I think it is ok.  The apps themselves come in as a dictionary from yaml, which has the same ordering issue right from the get go.  Having position at least gives some control over and we can add a feature where we have a way via settings on desktop to create an explicit ordering, which will probably be what we do if clients do start making their own apps or we end up with a bunch of panel apps.  Probably overkill right now.

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -48,6 +49,57 @@ def destroy_engine(self):
...
+        # logic
+        apps_tab = {"callback":   self.desktop_window._register_apps_tab,
+                    "properties": {"position": 0}}
+
+        panels = [apps_tab] + self._engine.panels.values()
+
+        def panel_position(panel):
+            """ Returns the relative position of a panel """
+            # place the panels with unspecified positions at the end
+            LAST_POSITION = float("inf")
```
jesseemond-adsk: any ideas to make this cleaner?

robblau: I can think of ways to make it shorter, but at the expense of clarity.  This seems just fine to me.

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -48,6 +49,57 @@ def destroy_engine(self):
...
+        :param title:        The title of the panel, to show as the tab name.
+        :param bundle:       The app, engine or framework object that is
+                             associated with this window.
+        :param widget_class: The class of the UI to be constructed. This must
+                             derive from QWidget.
+
+        Additional parameters specified will be passed through to the
+        widget_class constructor.
+        """
+        widget = widget_class(*args, **kwargs)
```
manneohrstrom: maybe we could add a debug printout here? Just to aid future troubleshooting.

jesseemond-adsk: would this do?

manneohrstrom:
This is looking awesome! One minor question: If I am reading the code right, it looks like any app that is included in the desktop engine which has a register_panel() call, will automatically be loaded into the engine.

This is cool and I guess you could describe this as being akin to for example nuke starting up, restoring a saved UI layout and as part of that process creating a shotgun panel. With the exception that in the desktop, you are not restoring a saved configuration, you are just launching all the panels that can be found.

Each panel knows its preferred position via the `position` property. 

I am wondering if it may make sense to try to separate the information used to tell the system about the panel from the information used to place it. This approach would be more akin to how other engines work. For example, in nuke or maya, we tell the system what to place in the 'favouries' area of the menu:

(from the config)
```
menu_favourites:
    - {app_instance: tk-multi-workfiles, name: Shotgun File Manager...}
    - {app_instance: tk-multi-snapshot, name: Snapshot...}
    - {app_instance: tk-multi-workfiles, name: Shotgun Save As...}
    - {app_instance: tk-multi-publish, name: Publish...}
```

In flame, we now have the same thing for commands running at startup:

```
run_at_startup:
    - {app_instance: tk-multi-workfiles, name: Shotgun File Manager...}
    - {app_instance: tk-multi-snapshot, name: Snapshot...}
    - {app_instance: tk-multi-workfiles, name: Shotgun Save As...}
    - {app_instance: tk-multi-publish, name: Publish...}
```

I think if we introduced something like that for the desktop, this would be more in line with the rest of the engines - for example:

```
panels:
    - {app_instance: tk-multi-workfiles, name: Shotgun File Manager...}
    - {app_instance: tk-multi-snapshot, name: Snapshot...}
    - {app_instance: tk-multi-workfiles, name: Shotgun Save As...}
    - {app_instance: tk-multi-publish, name: Publish...}
```

Would this make sense conceptually? The positioning attribute wouldn't be needed, the order of the list would control that. Each of the registered commands callbacks defined in the `panels` setting above would need to launch and return a panel object but i think this is already a requirement (here: https://github.com/shotgunsoftware/tk-desktop/pull/25/files#diff-b96c46b4274b8838e9ce3ff8d0a5ab38R81 )

Curious to hear what you think. Definitely want to keep things simple but also nice if we can keep things aligned with how other engines work - this may make future actions work easier in the future. For example, imagine a workflow where we may want to turn on and off panels dynamically via menu items.

This is totally awesome work btw. Super super stoked about this!! :-)

jesseemond-adsk:
@manneohrstrom I would actually prefer that approach. Specifying `position` in the panel registration options feels odd and auto-registration reduces flexibility.

Would it make sense to change the `panels` property name in the config to `tabs` instead? I personally would find it more intuitive, although maybe that is due to my lack of experience with toolkit in general. It would look like this:
   
    panels:
        - {app_instance: tk-multi-workfiles, name: Shotgun File Manager...}
        - {app_instance: tk-multi-snapshot, name: Snapshot...}
        - {app_instance: tk-multi-workfiles, name: Shotgun Save As...}
        - {app_instance: tk-multi-publish, name: Publish...}

Also, relatively minor issue: we wouldn't be able to specify tabs that appear *before* `Apps`, or at least not until we move the `Apps` functionality to an app on its own. Is that a problem?

manneohrstrom:
`tabs` sounds great! Let's do it!

For the `Apps` bit, maybe we should just wait and let it play out? You had some ideas of making the "springboard" part of the desktop (as I always used to think of it, like the launcher on the iPhone :) ) an actual app - in that case it would be naturally ordered like any other app. Alternatively, if we keep it bundled with the desktop, I guess we could always introduce some kind of special/system/built-in type of entry to cover it and allow it to be positioned alongside other tabs. This could be done as an addition, in a backwards compatible way, so i would probably suggest we take a wait and see approach for this one - what do you think?

robblau:
Because this isn't being merged directly to master I asked Jesse to create a ticket for the tabs setting so we could tackle this there.  We know that we want Apps to become Projects and move after My Tasks, so we need something that will work with that.  Once we get further into My Tasks we'll have a better idea what it would take to make Projects its own app and can figure this out then.

I'd also like to know if we could avoid copying the code that resolves settings like that into another spot... maybe find a home in core for it before we implement this functionality (since it really isn't the most intuitive bit of config).

PR #26, opened 2015-10-07
#26210 Change Tray Icon to Be Visible in Dark Mode

jesseemond-adsk:
Change the tray icon to be blue so that it is visible while in "Dark mode" on OSX.

Ideally, it would have been nice to use [`setTemplate`](http://indiestack.com/2014/10/yosemites-dark-mode/), but that was [only recently added to Qt](https://codereview.qt-project.org/#/c/115120/), so it is hardly an option (not even part of a Qt release as of this writing).

If we really want to follow the OS convention, we can revisit this issue later on to [dynamically detect if the OS is in "Dark mode"](http://stackoverflow.com/a/25214873) and change the tray icon, but that would require a little bit more involvement.

(**no longer how it looks, see below for the updated icon**)
Here is how it looks in normal mode now:
screen shot 2015-10-07 at 4 37 43 pm

And in "Dark mode":
screen shot 2015-10-07 at 4 37 52 pm


jesseemond-adsk:
Adapted icon looks like this:
screen shot 2015-10-08 at 9 46 16 am

Dark:
screen shot 2015-10-08 at 9 46 30 am

PR #27, opened 2015-10-13
Ticket #31606 - Update Desktop to respect project specific debug flag

robblau:
Make it so that if the project specific python emits a line from logging the Desktop will show it, even if the message is a debug message and the site config is not in debug mode.  This way the settings at the project level are respected.

### DIFF
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -370,7 +370,8 @@ class DesktopEngineSiteImplementation(object):
         self._engine._logger.log(level, msg, *args)
 
     def proxy_log(self, level, msg, args):
-        self._engine._logger.log(level, "[PROXY] %s" % msg, *args)
+        # Use a high level so that messages we are a proxy for will always be logged
+        self._engine._logger.log(100, "[PROXY %s] %s" % (logging.getLevelName(level), msg), *args)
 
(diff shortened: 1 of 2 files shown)


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -370,7 +370,8 @@ def log(self, level, msg, *args):
         self._engine._logger.log(level, msg, *args)
 
     def proxy_log(self, level, msg, args):
-        self._engine._logger.log(level, "[PROXY] %s" % msg, *args)
+        # Use a high level so that messages we are a proxy for will always be logged
+        self._engine._logger.log(100, "[PROXY %s] %s" % (logging.getLevelName(level), msg), *args)
```
manneohrstrom: Seems perhaps a little odd to escalate debug log messages (and others) to a more critical level in order to display them. What happens for example with color coding in the UI? Future changes may have unexpected results because we are elevating log message status.

To deal with this, could there instead be different logging channels? we could have a root logger and then each project and the site config would all have different loggers, each with different log levels. We would attach the log file and the log window to the top level logger. I think this approach is more along the lines of the python logging system designs and intentions. But maybe it's complicated - i don't know enough about the desktop design to have a good understanding of that.

jfboismenu:
Closing this, as we've totally revamped our logging in the last few years.

PR #28, opened 2015-10-28
#31557 Friendlier Errors on Empty OS Path

JesseEmond:
Two small fixes with this PR:

- Friendlier error message if desktop tries to open a pipeline config that doesn't have a path set for the current OS.
- Allow to change pipeline config even if the current pipeline config fails because of an empty OS path (useful as a dev if e.g. Windows is not set on primary, but it is set on an other pipeline config for testing purposes). This is done by moving the context menu initialization before the code that reads the config.

Here's how master currently handles it:

screen shot 2015-10-27 at 3 14 56 pm

Two things to notice here:

- error message is too technical
- can't change primary config

With this fix:
screen shot 2015-10-27 at 3 15 49 pm

### DIFF
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -710,6 +710,9 @@ class DesktopWindow(SystrayWindow):
                         "falling back to primary." % pipeline_configuration_id)
                     pipeline_configuration = primary_pipeline_configuration
 
+            # going to launch the configuration, update the project menu if needed
+            self.__populate_pipeline_configurations_menu(pipeline_configurations, pipeline_configuration)
+
             config_path = pipeline_configuration[path_field]
 
             # Now find out the appropriate python to launch
@@ -771,9 +774,6 @@ class DesktopWindow(SystrayWindow):
                 "\n\n%s\n\nSee the console for more details." % str(error)
             self.project_overlay.show_error_message(message)
             return
-
-        # going to launch the configuration, update the project menu if needed
-        self.__populate_pipeline_configurations_menu(pipeline_configurations, pipeline_configuration)
 
         core_python = os.path.join(core_root, "install", "core", "python")
 
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -722,6 +722,13 @@ class DesktopWindow(SystrayWindow):
             else:
                 raise RuntimeError("unknown platform: %s" % sys.platform)
 
+            if config_path == None:
+                raise RuntimeError("No path set for %s on the Pipeline "
+                                   "Configuration id %d. Please make sure that "
+                                   "it is set on your Shotgun website." %
+                                   (current_platform,
+                                    pipeline_configuration_id))
+
             current_config_path = config_path
             while True:
                 # First see if we have a local configuration for which interpreter
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -725,7 +725,7 @@ class DesktopWindow(SystrayWindow):
             else:
                 raise RuntimeError("unknown platform: %s" % sys.platform)
 
-            if config_path == None:
+            if config_path is None:
                 raise RuntimeError("No path set for %s on the Pipeline "
                                    "Configuration id %d. Please make sure that "
                                    "it is set on your Shotgun website." %
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -723,16 +723,18 @@ class DesktopWindow(SystrayWindow):
             elif sys.platform.startswith("linux"):
                 current_platform = "Linux"
             else:
-                raise RuntimeError("unknown platform: %s" % sys.platform)
+                raise RuntimeError("Unknown platform: %s." % sys.platform)
 
             if config_path is None:
-                raise RuntimeError("No path set for %s on the Pipeline "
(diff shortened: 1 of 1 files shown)


python/tk_desktop/desktop_window.py
```diff
@@ -722,6 +725,13 @@ def __launch_app_proxy_for_project(self, project, pipeline_configuration_id=None
             else:
                 raise RuntimeError("unknown platform: %s" % sys.platform)
 
+            if config_path is None:
+                raise RuntimeError("No path set for %s on the Pipeline "
+                                   "Configuration id %d. Please make sure that "
+                                   "it is set on your Shotgun website." %
```
robblau: Should add a break after the first sentence so that the error fits better in the area we have.
Add in the name of the pipeline configuration in addition to the id so it is easier to find in SG.

JesseEmond: Something like this?

screen shot 2015-10-29 at 11 10 17 am

robblau: Looks good.

jfboismenu-adsk: Should we also mention install_location.yml. Sometimes people edit those things manually.

manneohrstrom:
Can we call the OSes Mac, Windows and Linux in the message? Most people don't know what Darwin is.

manneohrstrom:
A few more comments and thoughts:

- It looks a little odd that the two of the three sentences end with a period. I think it should be consistent. I suggest adding a period to the first sentence. 

- Also, what does "Error setting up engine environment actually mean?" It doesn't make much sense to me in this context.

- Also do we know that this message will *only* be shown when the OS path is not set in shotgun? What about other cases? Thinking about the case of a missing or out of sync install location (like JF says). 

- Technically, updating the pipeline configuration path in Shotgun is currently not enough. You also need to do update the `install_location.yml` file with the path. The correct approach right now (confusing I know!) is to actually run the `tank move_configuration` command to resolve this issue.

- I think we should tackle the above confusion as a separate ticket to make sure that if you update the PC path Shotgun, it automatically updates the path in `install_location.yml` if needed. This means we can provide a message here which is similar to the one Jesse is suggesting.

Given the above, I think it would be more helpful to say something like:

<hr>
The Toolkit configuration path has not been set 
for your operating system.

To resolve this, please navigate to this project in Shotgun
and check the path for the Pipeline Configuration.

For more details, see the console.
<hr>

I think we should loop in @kporangehat to make sure he is happy with whatever copy with end up with.

The case when the pipeline config isn't `primary` is unusual and will typically only be hit by
technical users, so I suggest putting details relating to that (pipeline config id and name etc) in the console rather than in the main message.

JesseEmond:
I tried to make the errors more user-friendly with my previous commit. The technical details are showed in the console instead. Here are a few scenarios that I have tested (left is the console, right is the desktop window):

- empty path on the pipeline config, correct path in `install_location.yml` (doesn't really matter here):
screen shot 2015-11-02 at 10 48 02 am

- *invalid* path on the pipeline config, correct path in `install_location.yml` (doesn't really matter here):
screen shot 2015-11-02 at 10 49 10 am

- correct path on the pipeline config, *empty* path in `install_location.yml` (not handled by this PR - I believe that would touch tk-core instead):
screen shot 2015-11-02 at 10 50 34 am

- correct path on the pipeline config, *invalid* path in `install_location.yml` (idem, not handled by this PR, that's in core):
screen shot 2015-11-02 at 10 51 04 am

- sanity check with both config paths properly set:
screen shot 2015-11-02 at 10 58 15 am


I didn't bother with `Darwin` v.s. `Mac`, as I felt this would require touching the code (+ ~3 lines?) only to get the message changed in the console. I felt like the console could do with that small technicality, but feel free to tell me if you disagree.

manneohrstrom:
Loving it!!! Curious what @kporangehat thinks! Nice work Sir!

kporangehat:
This is trending in a really nice direction!

@JesseEmond In the 2nd example:

> invalid path on the pipeline config, correct path in install_location.yml (doesn't really matter here):

Wouldn't we want to direct the user to update this info in Shotgun again as we did in example #1?

```
To resolve this, please navigate to this project in Shotgun 
and check the path for the Pipeline Configuration.
```

Also, do we have the PipelineConfiguration `id` at this point as well as the Shotgun server url? It would be *really* nice to provide a link that would take them right there no?

---

To resolve this, [open Shotgun in your browser](https://kp.shotgunstudio.com/detail/PipelineConfiguration/141) and check the paths for this Pipeline Configuration.

---

Though maybe on a detail page this info isn't as obvious as on the list view for all PipelineConfigurations in a Project. Which we don't have any way to get a url for... :-1:

JesseEmond:
@kporangehat that actually sounded like a *very* good idea, although the errors in this case use `tk-framework-qtwidgets`' overlay for the error text, which is simply **painting** the text. This prevents us from adding rich text (e.g. links), and putting the raw URL there doesn't help either because it can't be selected or copied... I tried it and it was more frustrating than anything to have an unclickable/unselectable link there.

I have create an internal ticket (# 33179) to allow `tk-framework-qtwidgets`'overlay to handle widgets.

I agreed with your feedback and I have changed the error messages so that both cases advise the user to go to the website:
screen shot 2015-11-06 at 2 02 12 pm
screen shot 2015-11-06 at 2 03 56 pm

Unfortunate that we can't put a link in there yet, but would it be good enough for now?

kporangehat:
looks good :moneybag:

PR #29, opened 2015-11-05
#31592 Register Tabs Through Config

JesseEmond:
Register tabs straight from the config rather than being automatically registered when added as an app (which was less flexible). For example, registering an application to be shown as a tab in desktop would now look like:

```
tabs:
    - {app_instance: tk-desktop-mytasks, name: ''}
```

This uses the new version of *tk-core*, which provides a `get_matching_commands` function to match the commands that match the tab selectors. Therefore, the core minimum version requirement was bumped up to **v0.16.33**.

Here is a tabs registration test with the results:
screen shot 2015-11-05 at 4 03 53 pm

A couple of things to note:
- **first registered tab** was properly registered as the first *My Tasks* tab. No errors there (it matched any command registered by *tk-desktop-mytasks* because of the empty string name selector `''`).
- **second registered tab** was properly registered as the second *My Tasks* tab. No errors there (it matched the command with the exact `'tk-desktop-mytasks2:My Tasks...'` name).
- **third registered tab** failed to register because `tk-bad-app` is not a loaded app. Warning properly shown in console.
- **fourth registered tab** failed to register because `'badname'` does not match any command registered by `tk-desktop-mytasks-badname` (even though the app was properly loaded). Warning properly shown in console.

### DIFF
--- a/info.yml
+++ b/info.yml
@@ -138,7 +138,7 @@ description: "The engine that runs inside the Shotgun Desktop Application"
 
 # Required minimum versions for this item to run
 requires_shotgun_version:
-requires_core_version: "v0.16.3"
+requires_core_version: "v0.16.33"
 
 frameworks:
     - {"name": "tk-framework-qtwidgets", "version": "v1.x.x"}
--- a/info.yml
+++ b/info.yml
@@ -18,6 +18,25 @@ configuration:
                      commands that do not match any of the values in the groups
                      setting."
 
+    tabs:
+        type: list
+        description: "Controls what apps will show up as tabs.  This is a list
+                     where each element is a dictionary with two keys:
+                     'app_instance' and 'title'. The app_instance value connects
+                     this entry to a particular app instance defined in the
+                     environment configuration file. The title is the string
+                     displayed to select that tab. The order in which the tabs
+                     appear in this list defines the order in which the tabs are
+                     displayed (first element being the leftmost tab, last
+                     element being the rightmost tab)."
+        allows_empty: True
+        default_value: []
+        values:
+            type: dict
+            items:
+                title: { type: str }
+                app_instance: { type: str }
+
     groups:
         type: list
         description: "A list of dictionaries that define what commands get put
--- a/info.yml
+++ b/info.yml
@@ -21,14 +21,14 @@ configuration:
     tabs:
         type: list
         description: "Controls what apps will show up as tabs.  This is a list
-                     where each element is a dictionary with three keys:
-                     'app_instance' and 'name'. The app_instance value connects
-                     this entry to a particular app instance defined in the
-                     environment configuration file. The name is the panel to
-                     run when the tabs are added to the UI. The order in which
-                     the tabs appear in this list defines the order in which the
-                     tabs are displayed (first element being the leftmost tab,
-                     last element being the rightmost tab)."
+                     where each element is a dictionary with two keys:
+                     'app_instance' and 'name'. The 'app_instance' value
+                     connects this entry to a particular app instance defined in
+                     the environment configuration file. 'name' is the command
+                     to run when the tabs are added to the UI. The order in
+                     which the tabs appear in this list defines the order in
+                     which the tabs are displayed (first element being the
+                     leftmost tab, last element being the rightmost tab)."
         allows_empty: True
         default_value: []
         values:
--- a/info.yml
+++ b/info.yml
@@ -30,7 +30,7 @@ configuration:
                      desktop starts up. If 'name' is '', then all commands from
                      the given app instance are started."
         allows_empty: True
-        default_value: []
+        default_value: [{app_instance: tk-desktop, name: ''}]
         values:
             type: dict
             items:
--- a/info.yml
+++ b/info.yml
@@ -21,20 +21,20 @@ configuration:
     tabs:
         type: list
         description: "Controls what apps will show up as tabs.  This is a list
-                     where each element is a dictionary with two keys:
-                     'app_instance' and 'title'. The app_instance value connects
+                     where each element is a dictionary with three keys:
+                     'app_instance' and 'name'. The app_instance value connects
                      this entry to a particular app instance defined in the
-                     environment configuration file. The title is the string
-                     displayed to select that tab. The order in which the tabs
-                     appear in this list defines the order in which the tabs are
-                     displayed (first element being the leftmost tab, last
-                     element being the rightmost tab)."
+                     environment configuration file. The name is the panel to
+                     run when the tabs are added to the UI. The order in which
+                     the tabs appear in this list defines the order in which the
+                     tabs are displayed (first element being the leftmost tab,
+                     last element being the rightmost tab)."
         allows_empty: True
         default_value: []
         values:
             type: dict
             items:
-                title: { type: str }
+                name: { type: str }
                 app_instance: { type: str }
 
     groups:
--- a/info.yml
+++ b/info.yml
@@ -35,7 +35,7 @@ configuration:
                        {app_instance: '', name: 'Apps'}
                      "
         allows_empty: True
-        default_value: [{app_instance: '', name: 'Apps'}]
+        default_value: [{app_instance: '', name: Apps}]
         values:
             type: dict
             items:
--- a/info.yml
+++ b/info.yml
@@ -28,9 +28,14 @@ configuration:
                      configuration file.
                      The 'name' is the menu name of the command to run when
                      desktop starts up. If 'name' is '', then all commands from
-                     the given app instance are started."
+                     the given app instance are started.
+
+                     To add the 'Apps' tab (which contains the app launcher),
+                     add the following startup command to the list:
+                       {app_instance: '', name: 'Apps'}
+                     "
         allows_empty: True
-        default_value: [{app_instance: tk-desktop, name: ''}]
+        default_value: [{app_instance: '', name: 'Apps'}]
         values:
             type: dict
             items:
--- a/info.yml
+++ b/info.yml
@@ -18,17 +18,17 @@ configuration:
                      commands that do not match any of the values in the groups
                      setting."
 
-    tabs:
+    run_at_startup:
         type: list
-        description: "Controls what apps will show up as tabs.  This is a list
+        description: "Controls what apps will run on startup.  This is a list
                      where each element is a dictionary with two keys:
-                     'app_instance' and 'name'. The 'app_instance' value
-                     connects this entry to a particular app instance defined in
-                     the environment configuration file. 'name' is the command
-                     to run when the tabs are added to the UI. The order in
-                     which the tabs appear in this list defines the order in
-                     which the tabs are displayed (first element being the
-                     leftmost tab, last element being the rightmost tab)."
+                     'app_instance' and 'name'.
+                     The 'app_instance' value connects this entry to a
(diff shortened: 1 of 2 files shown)


info.yml
```diff
@@ -18,6 +18,25 @@ configuration:
                      commands that do not match any of the values in the groups
                      setting."
 
+    tabs:
+        type: list
+        description: "Controls what apps will show up as tabs.  This is a list
+                     where each element is a dictionary with three keys:
```
jfboismenu: Is it two or three?

info.yml
```diff
@@ -18,6 +18,25 @@ configuration:
...
+                     'app_instance' and 'name'. The app_instance value connects
```
jfboismenu: 'app_instance'

info.yml
```diff
@@ -18,6 +18,25 @@ configuration:
...
+                     this entry to a particular app instance defined in the
+                     environment configuration file. The name is the panel to
```
jfboismenu: 'name'

info.yml
JesseEmond: artifact from my previous approach. Thanks!

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -57,28 +57,16 @@ def _init_tabs(self):
...
-
-            return panel["properties"].get("position", LAST_POSITION)
-
-        ordered_panels = sorted(panels, key=panel_position)
-
-        for panel in ordered_panels:
+        for (_,_,panel_callback) in commands:
             # let the panel show itself through its registered callback
-            panel["callback"]()
+            panel_callback()
```
manneohrstrom: There is a lot of stuff happening here - i think some further explanation in comments would add value.
I would add something along these lines:

```
# panel apps all register commands that when executed will request the engine 
# to show the panel ui. In something like maya or nuke, these commands would appear
# on the shotgun menu and be triggered by the user clicking on them.
#
# here, in desktop, all commands added to the `tabs` configuration setting will
# be executed, a panel is expected to be shown as part of the execution of the 
# panel and when the panel is shown, it is placed in the next available tab slot in
# the desktop, going from left to right.
```

manneohrstrom:
Looking at this now, I wonder if it is weird that we call this setting `tabs` - when what it really does is to simply execute things at startup. Maybe it should be more like flame - here: https://github.com/shotgunsoftware/tk-flame/blob/master/info.yml#L49

right now, if you would register a command that *doesn't* show a panel (say the loader), it would still work, right - just pop up the loader UI instead? This makes me think that maybe this is not `tabs`, it's actually `run_at_startup`. Things are executed in list order. Commands that launch panels will then basically mount the panel at that point. Panels are always mounted to the right, meaning that you could potentially execute `show_panel()` whenever you want, not necessarily just at startup. Imagine an app that launches as a dialog but has a "dock" button - which basically just executes show_panel() on that same UI (or something like that). That should simply create a panel and place it in the next panel slot in the desktop, to the right of all other panels.

The only potential issue with this is if we wanted "Apps" to appear say in a second slot - for example having the tab order `My Tasks | Apps | Something else`. I think if we wanted to do this, we could just refactor the Apps tab to be an app rather than a built-in thing.

Thoughts?

JesseEmond:
After your explanation of the commands workflow, I think that your suggestion makes more sense. I personally think it might look a bit odd for someone that is not familiar with toolkit to need to "register commands at startup to add tabs", but nothing that documentation can't explain. Additionally, it is consistent with other apps too. I will update my PR later today.

JesseEmond:
Modified to treat "Apps" as a fake registered application. This allows to change the tab's position (or even remove it!) as desired. I changed the logic to execute startup applications to do this:
- find the "Apps" selector in the config
- strip it off of the list of selectors (or else `get_matching_commands` will show warnings)
- get the matching commands
- figure out where to inject a "fake" command that will register the "Apps" tab
- execute the commands

Figuring out where to inject the "Apps" command is the trickiest part. Please let me know how I could improve that part of the code to clean it up... This would eventually replaced by moving "Apps" to an app on its own.  

Since this plays quite a bit with indices, I tested a couple of things to be sure.
Here are the tests that I have done (all had the expected behaviour):
- add as first tab like so: `- {app_instance: tk-desktop, name: 'Apps'}`
- add as first tab like so: `- {app_instance: tk-desktop, name: ''}`
- add as last tab (with no command-matching errors before):
```
- {app_instance: tk-desktop-mytasks, name: ''}
- {app_instance: tk-desktop-mytasks2, name: 'tk-desktop-mytasks2:My Tasks...'}
- {app_instance: tk-desktop, name: ''}
```
- add as middle tab (with no command-matching errors before):
```
- {app_instance: tk-desktop-mytasks, name: ''}
- {app_instance: tk-desktop, name: 'Apps'}
- {app_instance: tk-desktop-mytasks2, name: 'tk-desktop-mytasks2:My Tasks...'}
```
- add as middle tab, with command-matching errors *after*:
```
- {app_instance: tk-desktop-mytasks, name: ''}
- {app_instance: tk-desktop, name: 'Apps'}
- {app_instance: tk-desktop-mytasks2, name: 'tk-desktop-mytasks2:My Tasks...'}
- {app_instance: tk-bad-app, name: ''}
- {app_instance: tk-desktop-mytasks-badname, name: 'badname'}
```
- add as middle tab, with command-matching errors *before*, with command-matching successes before as well:
```
- {app_instance: tk-desktop-mytasks, name: ''}
- {app_instance: tk-bad-app, name: ''}
- {app_instance: tk-desktop-mytasks-badname, name: 'badname'}
- {app_instance: tk-desktop, name: 'Apps'}
- {app_instance: tk-desktop-mytasks2, name: 'tk-desktop-mytasks2:My Tasks...'}
```
- add as last tab, with command-matching errors *before* (no other):
```
- {app_instance: tk-bad-app, name: ''}
- {app_instance: tk-desktop-mytasks-badname, name: 'badname'}
- {app_instance: tk-desktop, name: 'Apps'}
```
- *no* "Apps" tab:
```
- {app_instance: tk-desktop-mytasks, name: ''}
- {app_instance: tk-desktop-mytasks2, name: 'tk-desktop-mytasks2:My Tasks...'}
```
- add as last tab, with app before that generate 3 tabs:
```
- {app_instance: tk-desktop-mytasks, name: ''}      (customized to spawn 3 panels)
- {app_instance: tk-desktop, name: 'Apps'}
``` 
- add as middle tab, with apps before and after that generate 3 tabs each:
```
- {app_instance: tk-desktop-mytasks, name: ''}      (customized to spawn 3 panels)
- {app_instance: tk-desktop, name: 'Apps'}
- {app_instance: tk-desktop-mytasks2, name: ''}      (customized to spawn 3 panels)
```
- add as last tab, with apps before that generate 3 tabs and an app that has a command-matching error before:
```
    - {app_instance: tk-desktop-mytasks, name: ''}       (customized to spawn 3 panels)
    - {app_instance: tk-bad-app, name: ''}
    - {app_instance: tk-desktop, name: 'Apps'}
```

info.yml
```diff
@@ -18,6 +18,25 @@ configuration:
...
+        type: list
+        description: "Controls what apps will run on startup.  This is a list
+                     where each element is a dictionary with two keys:
+                     'app_instance' and 'name'.
+                     The 'app_instance' value connects this entry to a
+                     particular app instance defined in the environment
+                     configuration file.
+                     The 'name' is the menu name of the command to run when
+                     desktop starts up. If 'name' is '', then all commands from
+                     the given app instance are started."
```
robblau: Need to document how to control where the "Apps" tab lives.

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -52,42 +52,82 @@ def destroy_engine(self):
...
+        This gives an opportunity for apps to display panels through the
+        execution of a command, which will add a tab to Desktop.
+        """
+        selectors = self._engine.get_setting("run_at_startup", [])
+
+        # "Apps" is currently a builtin command, so we first figure out where it
+        # is (its index) in the selectors:
+        def is_apps_selector(selector):
+            return (selector["app_instance"] == self._engine.instance_name and
+                    (selector["name"] == "" or selector["name"] == "Apps"))
```
robblau: Is there a reason to include an empty name as a valid name for Apps?  Seems like making it more specific would be better.  Also, maybe have the app_instance be empty for this?  That would be invalid elsewhere so there wouldn't be a conflict and it would make it so that if you renamed the engine instance name for any reason you wouldn't have to go in and update the setting...

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -52,42 +52,82 @@ def destroy_engine(self):
...
+        apps_index = next((i for i, selector in enumerate(selectors)
+                           if is_apps_selector(selector)),
+                          None)
```
robblau: If the above changes make sense and the descriptor is static then this could become 

selectors.index({"app_instance": "", "name": "Apps"})

Which reads a lot easier.

python/tk_desktop/desktop_engine_site_implementation.py
JesseEmond: Good point. I added the empty string to stay consistent with the other selectors, but it probably is easier and safer to just support one syntax. Empty `app_instance` sounds like a better idea too. Updating!

python/tk_desktop/desktop_engine_site_implementation.py
JesseEmond: Thanks for the idea! I find it much cleaner that way. Will change.

info.yml
```diff
@@ -18,6 +18,30 @@ configuration:
...
+                     'app_instance' and 'name'.
+                     The 'app_instance' value connects this entry to a
+                     particular app instance defined in the environment
+                     configuration file.
+                     The 'name' is the menu name of the command to run when
+                     desktop starts up. If 'name' is '', then all commands from
+                     the given app instance are started.
+
+                     To add the 'Apps' tab (which contains the app launcher),
+                     add the following startup command to the list:
```
robblau: Should mention where Apps goes if this isn't there... since this seems to imply that you could leave it out.

JesseEmond: @robblau Currently, leaving it out does not add the tab. So, if you have the following:

```
- {app_instance: tk-desktop-mytasks, name: ''}
- {app_instance: tk-desktop-mytasks2, name: 'tk-desktop-mytasks2:My Tasks...'}
``` 
No `Apps` tab will show up.

I figured that this would add free flexibility (not sure why you wouldn't want that tab, but it's possible to leave it out anyway) and the `default_value` would take care of the default case that would only use `Apps` and nothing else.

Do you think it would be better if we had a fallback behaviour when the `Apps` selector is not in the config? It felt a bit arbitrary to add this firsthand, but I can see why it could be nice to have the tab added no matter what.

robblau: Nope, that would make sense (I'm all for flexibility).  But in that case there should be a default value for this that does the right thing.  Just in case somebody ends up using add_engine to add tk-desktop to an environment.

JesseEmond: @robblau Oops, you are right! Forgot about `get_setting`'s default value and thought you meant `info.yml`'s `default_value`. Updated! (not sure how to quickly test the `add_engine` use-case though - does the code look right to you?)

JesseEmond:
Merged.

PR #30, opened 2016-01-12
#31598 Added support for cloud based config sync

manneohrstrom:
This adds a up-to-date check to the desktop. When a user clicks a project, the desktop now will fall back on a cloud based config check in the case no paths have been defined in the pipeline configuration.

The extra check doesn't affect the performance of Desktop. If a zip config is available in Shotgun, the new `tank synchronize` command is called and its progress reporting is displayed in the desktop window as it potentially updates the config, downloads apps, engines etc.

**NOTE:** The `PIPELINE_CONFIG_ATTACHMENT_FIELD` is likely to change before an official release.


manneohrstrom:
deleting old spike PR

PR #32, opened 2016-03-08
possibility to use environment variables in share core config files

adriankrupa:
~~apparently it breaks browser integration, don't know how to fix it~~
fixed also [here](https://github.com/shotgunsoftware/tk-core/pull/250) and [here](https://github.com/shotgunsoftware/tk-framework-desktopstartup/issues/23)


adriankrupa:
apparently it breaks browser integration, don't know how to fix it

PR #33, opened 2016-05-23
For #29128: Don't use the event log for tracking recent activity

kporangehat:
Previously we used the event log to help track recent app launches and recently accessed projects per user. This was causing performance issues due to the additional queries required to calculate this information (especially on the event log which can be very slow).

With this change, we now track recent app launches using the [settings module in the Shotgun Utils Framework](http://developer.shotgunsoftware.com/tk-framework-shotgunutils/settings.html). These are tracked for the user on a per-project level and are unique to each machine the user is using. Recently accessed Projects are now solely tracked by the `Project.last_accessed_by_currrent_user` field on the Project in Shotgun. This information is updated whenever the Project is accessed via SG Desktop as well.

### DIFF
--- a/python/tk_desktop/project_commands_model.py
+++ b/python/tk_desktop/project_commands_model.py
@@ -9,6 +9,8 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 import datetime
+import time
+import re
 
 from sgtk.platform.qt import QtCore, QtGui
 
@@ -104,6 +106,8 @@ class ProjectCommandProxyModel(GroupingProxyModel):
 
 
 class ProjectCommandModel(GroupingModel):
+    APP_LAUNCH_EVENT_TYPE = "Toolkit_Desktop_AppLaunch"
+
     RECENT_GROUP_NAME = "RECENT"
 
     BUTTON_NAME_ROLE = QtCore.Qt.UserRole + 1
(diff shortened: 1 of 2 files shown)


python/tk_desktop/project_commands_model.py
```diff
@@ -306,52 +286,29 @@ def _handle_command_triggered(self, item, command_name=None, button_name=None,
...
-                existing_info = self.__recents.setdefault(
-                    name, {"timestamp": time_stamp, "added": False})
-                if existing_info["timestamp"] < time_stamp:
-                    self.__recents[name]["timestamp"] = time_stamp
+    def __load_recents_from_settings(self):
+        """
+        Loads the dict of recently launched apps from the user settings. See above for the
+        format.
+        """
+        self.__recents = self._settings_manager.retrieve("sg_desktop.recent_apps", {},
```
robblau: Would it be possible to fall back to the EventLog lookup in the event that the settings aren't there?  At least that way the recents listing won't disappear on people after this upgrade, but it'll still be fast from the first time on.

kporangehat: So whenever someone clicks into a project for the first time (including one they just setup), and/or hasn't ever launched anything, it will do an event log lookup?

kporangehat:
We now fall back on querying the event log one time if the settings don't exist yet. This allows us to preserve the history of app launches from the event log.

python/tk_desktop/project_commands_model.py
manneohrstrom: I was hoping we could get rid of all event log reads once and for all. What is the consideration here? usability? IMHO I think we are potentially overcomplicating things here + adding an inconsistent access pattern where desktop sometimes calls shotgun, sometimes doesn't, which could maybe lead to weird errors for users? (e.g. internet is down - meaning i can access one project but not another, depending on the qsettings state for these projects).

kporangehat:
Note: @manneohrstrom's comment in the outdated diff section. In hindsight, I think I actually agree that the complication isn't really worth the usability gain. Though it may be a bit confusing to see your recents disappear, this does simplify things a lot and does get us completely out of the event log. But I'll adjust to however the team decides is the best way forward.

PR #34, opened 2016-05-27
for #36693: ensure we explicitly close os-level file descriptors

kporangehat:
OS-level file handles opened with `mkstemp()` don't get closed automatically and over time can contribute to reaching the OS limit of open file handles. Reaching the OS limit will generate errors like:

```
IOError: [Errno 24] Too many open files: '/tmp/tmpJ6g4Ke'
```

```
Windows Error [Error 32]
```

### DIFF
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -111,7 +111,7 @@ class DesktopWindow(SystrayWindow):
         self._current_user_id = user["id"]
         thumbnail_url = current_user.get("image")
         if thumbnail_url is not None:
-            (_, thumbnail_file) = tempfile.mkstemp(suffix=".jpg")
+            (fd, thumbnail_file) = tempfile.mkstemp(suffix=".jpg")
             try:
                 shotgun.download_url(connection, thumbnail_url, thumbnail_file)
                 pixmap = QtGui.QPixmap(thumbnail_file)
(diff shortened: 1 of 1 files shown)


python/tk_desktop/desktop_window.py
```diff
@@ -799,8 +800,9 @@ def __launch_app_proxy_for_project(self, project, pipeline_configuration_id=None
                 "proxy_auth": server_auth,
             },
         }
-        (_, pickle_data_file) = tempfile.mkstemp(suffix='.pkl')
+        (fd, pickle_data_file) = tempfile.mkstemp(suffix='.pkl')
         pickle.dump(desktop_data, open(pickle_data_file, "wb"))
```
jfboismenu: Should `try`/`except` around dump, just in case. Otherwise this is good to go.

kporangehat: I didn't modify that section so I wasn't looking to change anything else but the file handle issue. If you want me to also change that to add a `try/except` I will. Just trying to keep this focused on one issue.

jfboismenu: Well, in this branch you are worried that file descriptors will be leaked and this is a potential leak, so it would make sense to clean that up as well. :)

jfboismenu:
Fixed in the Python 3 branch.

PR #36, opened 2016-06-23
For #36848, desktop with core v0.18.0

jfboismenu:
This switches the desktop engine code to use the 0.18 logging framework. It also bumps the minimal version of core required to run the desktop engine. I also removed the unused Shotgun API that was bundled.


python/shotgun_api3/lib/httplib2/__init__.py
```diff
@@ -1,1514 +0,0 @@
-from __future__ import generators
```
jfboismenu: Everything from here on out is shotgun api being removed, so no need to scroll further down.

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -206,12 +207,6 @@ def run(self, splash, version, **kwargs):
         # this engine.
         self.startup_version = kwargs.get("startup_version")
 
-        server = kwargs.get("server")
```
jfboismenu: This will need to be fixed by an update to the websocket server unfortunately.

manneohrstrom:
👍 !

PR #37, opened 2016-07-19
For #31490, background python leaks

jfboismenu:
This branch fixes cleanup issues when the background process exits.

- The client/server code has been refactored into SiteCommunication/ProjectCommunication/CommunicationBase classes not only to make sense of the code, but also to make the pattern of starting up and shutting down behave the same regardless of how it is being invoked. Doing so fixed most of the hanging Python processes issues.

- Stepping out of the project while a dialog is active also doesn't leave a Python running in the background anymore.

- Fixed a weird Qt bug for processes without a presence in the taskbar. It meant that the first dialog of the background process would never be shown, giving the user the impression that a GUI based command wasn't launching.


engine.py
```diff
@@ -39,6 +39,18 @@ def init_engine(self):
         else:
             self._logger.setLevel(logging.INFO)
 
+        # Figure out which implementation we will use.  If the tk instance
```
jfboismenu: Need to do this check beforehand so that _define_qt_base, which can't be specialized inside the engine's implementation class, can add a workaround for a Windows dialog visibility bug.

engine.py
```diff
@@ -186,86 +193,75 @@ def __getattr__(self, name):
...
-                base["dialog_base"] = ProxyDialogPySide
-                self.log_debug("Successfully initialized PySide '%s' located in %s."
-                               % (PySide.__version__, PySide.__file__))
-                self._has_ui = True
-            except ImportError:
-                pass
-            except Exception, e:
-                self.log_warning("Error setting up pyside. Pyside based UI support will not "
-                                 "be available: %s" % e)
+        try:
```
jfboismenu: Reduces code complexity.

hooks/launch_python.py
```diff
@@ -35,7 +35,7 @@ def execute(self, project_python, pickle_data_path, utilities_module_path):
 
         # run hidden on windows
         startupinfo = None
-        if sys.platform == "win32":
+        if sys.platform == "win32" and not os.environ.get("SGTK_DESKTOP_BACKGROUND_CONSOLE"):
```
jfboismenu: Added for debugging when logging is not enough.

python/tk_desktop/desktop_engine_project_implementation.py
```diff
@@ -72,67 +113,111 @@ def post_app_init(self):
...
             else:
                 self._engine.log_debug("Quitting on disconnect")
                 app.quit()
         else:
-            # just close down the message thread otherwise
-            self._engine.log_debug("Disconnect with no UI.  Shutting down")
-            self.msg_server.close()
+            self._project_comm.shut_down()
+
+    def start_app(self):
```
jfboismenu: This was refactored from the bootstrap_utilities so init and teardown live side by side.

python/utils/bootstrap_utilities.py
```diff
@@ -43,39 +43,7 @@ def start_engine(data):
 
 def start_app(engine):
     """ Run the QApplication for the given tk-desktop engine """
-    if engine.has_ui:
```
jfboismenu: Moved into desktop's project engine implementation.

PR #39, opened 2016-08-25
For #37142, update framework dependencies

jfboismenu:
Updates the framework dependencies to the latest major version.

### DIFF
--- a/python/tk_desktop/grouping_list_view.py
+++ b/python/tk_desktop/grouping_list_view.py
@@ -18,7 +18,7 @@ from .ui import resources_rc
 from .grouping_model import GroupingModel
 from .action_list_view import ActionListView
 
-shotgun_view = sgtk.platform.import_framework("tk-framework-qtwidgets", "shotgun_view")
+views = sgtk.platform.import_framework("tk-framework-qtwidgets", "views")
 
 
 class DefaultGroupingHeader(QtGui.QPushButton):
@@ -50,7 +50,7 @@ class DefaultGroupingHeader(QtGui.QPushButton):
             self.setIcon(self.right_arrow)
 
 
-class DefaultGroupingHeaderDelegate(shotgun_view.WidgetDelegate):
+class DefaultGroupingHeaderDelegate(views.WidgetDelegate):
     """ Default delegate for header items """
     def _create_widget(self, parent):
         widget = DefaultGroupingHeader(parent)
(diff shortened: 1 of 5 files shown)


python/tk_desktop/grouping_list_view.py
```diff
@@ -114,7 +114,7 @@ class GroupingListView(ActionListView):
     expanded_changed = QtCore.Signal(str, bool)
 
     def __init__(self, parent=None):
-        QtGui.QListView.__init__(self, parent)
+        ActionListView.__init__(self, parent)
```
josh-t: personal preference, but maybe use `super()` instead?

python/tk_desktop/project_commands_widget.py
```diff
@@ -54,11 +54,10 @@
 """
 
 
-class AbstractCommandDelegate(shotgun_view.WidgetDelegate):
+class AbstractCommandDelegate(views.EditSelectedWidgetDelegate):
     def __init__(self, view):
-        self.view = view
         view.entered.connect(self._handle_entered)
-        shotgun_view.WidgetDelegate.__init__(self, view)
+        views.EditSelectedWidgetDelegate.__init__(self, view)
```
josh-t: same (`super()`)

python/tk_desktop/project_delegate.py
```diff
@@ -16,16 +16,16 @@
...
 ShotgunModel = shotgun_model.ShotgunModel
 
 
-class SgProjectDelegate(shotgun_view.WidgetDelegate):
+class SgProjectDelegate(views.EditSelectedWidgetDelegate):
     def __init__(self, view, size):
         self._size = size
         self._view = view
-        shotgun_view.WidgetDelegate.__init__(self, view)
+        views.EditSelectedWidgetDelegate.__init__(self, view)
```
josh-t: another if you end up changing the others.

python/tk_desktop/grouping_list_view.py
jfboismenu: I feel like we're in an habit of always calling the base class directly, especially when dealing with Qt based classes. From my understanding super is useful when dealing with multiple inheritance, which is not the case here.

josh-t: Understood. My personal preference is to use `super` because I don't have to go back and change all these lines when/if the base class changes. No biggie. :)

jfboismenu: True dat. Wonder why we got into that habit in the first place. You'd think that we would be using super to ease maintenance...

PR #41, opened 2016-09-02
For #38035, non string resiliency for logging

jfboismenu:


engine.py
```diff
@@ -158,8 +158,11 @@ def log_exception(self, msg, *args):
             # implementation has not been setup yet, log directly to the logger
             self._logger.exception(msg, *args)
         else:
-            # implementation has been setup, let it handle the logging
-            exception_msg = msg + "\n" + traceback.format_exc()
+            if msg is not None:
```
josh-t: nitpicky: swap the order to simplify slightly:

```
if msg:
    ...
else:
    ...
```

jfboismenu: This is not the same tough. If the user passes in 0 or False, the values won't get printed. Highly unlikely, but still.

PR #43, opened 2016-09-22
For #38056, refresh project fails when inside project

jfboismenu:
There is no project list to refresh when inside a project so we'll hide the menu item.

### DIFF
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -468,6 +468,9 @@ class DesktopWindow(SystrayWindow):
         self._project_proxy.sort(0)
 
     def _on_back_to_projects_clicked(self):
+        """
+        Invoked when the user leaves a project.
+        """
         engine = sgtk.platform.current_engine()
         engine.site_comm.shut_down()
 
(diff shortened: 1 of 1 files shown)


python/tk_desktop/desktop_window.py
```diff
@@ -481,6 +481,9 @@ def _on_back_to_projects_clicked(self):
         self.current_project = None
         self._save_setting("project_id", 0, site_specific=True)
 
+        # We'll be showing the project list, show back the menu item.
```
josh-t: minor: maybe reword this comment a bit?

python/tk_desktop/desktop_window.py
```diff
@@ -636,6 +639,10 @@ def __launch_app_proxy_for_project(self, project, pipeline_configuration_id=None
             engine = sgtk.platform.current_engine()
             engine.log_debug("launching app proxy for project: %s" % project)
 
+            # Always hide the Refresh Projects menu items when launching a Python for a project,
```
josh-t: minor: `launching a Python` - reword?

PR #44, opened 2016-11-10
#38647 No longer need to bundle/load fonts. Core will do this now.

josh-t:
This change is a part of the changes to `tk-core` represented here: https://github.com/shotgunsoftware/tk-core/pull/355

Now that core bundles and loads **Open Sans** it is no longer necessary for desktop to do so. Because desktop uses `_initialize_dark_look_and_feel`, the font will be loaded automatically. The engine already uses **Open Sans** in its stylesheet, so nothing else is required.


jfboismenu:
This should at the minimum have a core requirement on 0.18, or people running older cores will upgrade and not have the right font.

jfboismenu:
Could you hold off on this release as well? This might make my life easier if I can assume that all tk-desktop engines that depend on 0.18 support the new logging?

josh-t:
As per @robblau, we'll hold off on removing these for now given the standard transition plan. I've created the internal ticket to do this next year.

PR #47, opened 2016-12-17
For #39630, UI to display when no commands are found for a project.

akhronek:
Includes a link to a Shotgun Support article describing how to configure Software entities on a site.

Ready for initial review, some specific implementation comments have been added.


python/tk_desktop/desktop_engine_project_implementation.py
```diff
@@ -131,6 +134,13 @@ def _register_commands(self):
                 time.sleep(0.5)
 
             self._project_comm.call("trigger_register_command", name, gui_properties, groups)
+            project_command = {"name": name}
+            project_command.update(gui_properties)
+            project_commands.append(project_command)
+
+        # Let the proxy know command registration is complete and which
+        # commands were registered for this project.
+        self._project_comm.call_no_response("project_commands_finished", project_commands)
```
akhronek: Passing the list of commands that were registered because I couldn't figure out a reliable way to determine what was added from the desktop_window widget.  If there is in fact a way to figure out that information from the desktop_widget, then this isn't necessary.

resources/build_resources.sh
```diff
@@ -11,7 +11,7 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 # The path to where the PySide binaries are installed
-PYTHON_BASE="/Applications/Shotgun.app/Contents/Frameworks/Python"
+PYTHON_BASE="/Applications/Shotgun.app/Contents/Resources/Python"
```
akhronek: Did the PySide executables move in a recent Desktop installation?  The only thing under the Frameworks/Python/bin directory for me was a "python" link back to the Resources folder:
```
66:/Applications/Shotgun.app/Contents/Frameworks/Python/bin ls -l
total 8
lrwxr-xr-x  1 macuser  staff  36 Dec 16 16:11 python@ -> ../../../Resources/Python/bin/python
```

This is for Desktop v1.3.6.

python/tk_desktop/install_apps.py
```diff
@@ -0,0 +1,250 @@
...
+        self._icon_layout_spacing = 5
+        self._icon_layout_stretch = 1
+        self._icon_column_count = 4
+        self._config_link_color = "#0000FF"
+        self._config_link_text = "here"
+
+        # Hide this widget by default.
+        self.setVisible(False)
+
+    def build(self, project=None):
```
manneohrstrom: can we clarify the method name? It feels too generic.

manneohrstrom: Regarding `project=None` - our current use case always assumes that we are displaying this screen for a particular project. is it OTT to support the `project=None` case? I think we should require a specific project to keep things simple for the time being. We have had issues with this sort of thing in other places, where the logic has gotten a bit out of hand. So I think we should skip the default value for the parameter and on the contrary always require a project id.

jfboismenu:
Could you include a screenshot?

info.yml
```diff
@@ -6,6 +6,21 @@
...
+    # NOTE: The 'sg_software_entity' setting is only temporary and
+    #       will be removed once the Software entity becomes native
+    #       in Shotgun
+    sg_software_entity:
+        type: str
+        description: Shotgun CustomNonProjectEntity designated as the
+                     Software entity until it becomes native.
+        default_value: CustomNonProjectEntity23
+
+    software_entity_config_link:
```
jfboismenu: This would probably be better handled by a constant in the code. It probably wouldn't be something that a TD would want to configure.

python/tk_desktop/desktop_engine_project_implementation.py
jfboismenu: [This](https://github.com/shotgunsoftware/tk-desktop/blob/master/python/tk_desktop/desktop_engine_site_implementation.py#L139) knows we are adding into the project menu and [that](https://github.com/shotgunsoftware/tk-desktop/blob/master/python/tk_desktop/desktop_engine_site_implementation.py#L160-L162) is violating the encapsulation of the desktop window, which is where things get added. In the latter's case, you could add a method that sets a flag or something letting the dialog know that something was indeed added in addition to adding the element. Then when you call `project_commands_finished`.

jfboismenu: Also, you will need to add code that either allows to catch if an RPC call doesn't exist, as you are not guaranteed a client has the latest site engine version. You can 't simply try/except around `call_no_response`either, as you could get an exception because the server died or because the server performed some invalid operation. I would recommend to introduce a new exception type and raise it [here](https://github.com/shotgunsoftware/tk-desktop/blob/master/python/tk_desktop/rpc.py#L149). I think custom types pass through the RPC call, so the other class should be accessible from the caller side as well. At the minimum, you can test the exception's `__str__` to know if it 's an unknown call.

python/tk_desktop/desktop_window.py
```diff
@@ -411,6 +413,29 @@ def on_project_filesystem_folder_triggered(self):
...
+        # that were added to the context menu.
+        button_commands = []
+        menu_commands = []
+        for project_command in project_commands:
+            if project_command.get("type") == "context_menu":
+                menu_commands.append(project_command)
+            else:
+                button_commands.append(project_command)
+
+        if not button_commands:
```
jfboismenu: Here you would test for the flag I suggested earlier.

python/tk_desktop/install_apps.py
```diff
@@ -0,0 +1,250 @@
...
+        """
+        Query Shotgun for a list of Software images to display and
+        set the Software configuration link value
+
+        :param project: Current Shotgun Project instance (dict)
+        """
+        # Make sure we can get a handle to the current engine.
+        # This is primarily used for logging purposes.
+        engine = sgtk.platform.current_engine()
```
jfboismenu: Could you explain when this is possible? Since that piece of code is executed when the engine runs, I have a hard time understanding when this would happen. And if I have a hard time understanding, then it needs a comment. :p

python/tk_desktop/install_apps.py
```diff
@@ -0,0 +1,250 @@
...
+        self._clear_icons()
+
+        # Arrange the icons in a grid with self._icon_column_count
+        # columns and N rows based on the total number of icons
+        col_num = 0
+        icon_row_layout = None
+        last_software = len(sg_softwares) - 1
+        for i in range(len(sg_softwares)):
+            # Download the thumbnail from Shotgun
+            sg_icon = shotgun_data.ShotgunDataRetriever.download_thumbnail(
```
jfboismenu: Fixing the transparency is important, but I think the loading thumbnail can wait. We can polish this once we are sure everything else that's important has made it.

With that being said... visual thumbnail states (not loading, loading, loaded, error) are something I would LOOOOOOVE to see happen.

resources/build_resources.sh
jfboismenu: Yes, they moved a while back. I think you are first person in a long while that needed to update the UIs. Congrats!

info.yml
manneohrstrom: heads up that this is just temporary, will go away as soon as we have the official Software entity type in sg.

akhronek: For the software entity config link, the ticket specifically calls for an engine config setting.

python/tk_desktop/desktop_engine_project_implementation.py
```diff
@@ -132,6 +132,9 @@ def _register_commands(self):
 
             self._project_comm.call("trigger_register_command", name, gui_properties, groups)
 
+        # Let the proxy know command registration is complete
+        self._project_comm.call_no_response("project_commands_finished")
```
akhronek: Playing around with this, I found that the 'call_no_response' method actually swallows the ValueError raised when an unknown function is detected [here](https://github.com/shotgunsoftware/tk-desktop/blob/39630_no_item_overlay/python/tk_desktop/rpc.py#L166), so there is no need to have a try/except clause at this point in the code.  It would need to be handled here if I was using the 'call()' method, but I don't think that's necessary at this point.  Also, I tried implementing a custom error [here](https://github.com/shotgunsoftware/tk-desktop/blob/39630_no_item_overlay/python/tk_desktop/rpc.py#L149), but keep in mind that if the unknown function error is being raised in the first place, then it's probably because the site configuration needs to be updated, and the custom exception wouldn't be there anyway ;)  Does this all sound reasonable?

python/tk_desktop/no_apps_installed_overlay.py
```diff
@@ -0,0 +1,252 @@
...
+        # Arrange the icons in a grid with self._icon_column_count
+        # columns and N rows based on the total number of icons
+        col_num = 0
+        icon_row_layout = None
+        last_software = len(sg_softwares) - 1
+        for i in range(len(sg_softwares)):
+            # Download the thumbnail from Shotgun
+            sg_icon = shotgun_data.ShotgunDataRetriever.download_thumbnail(
+                sg_softwares[i]["image"], engine
+            )
```
akhronek: This will be updated to handle transparency as part of #39292.  Leaving as-is for now.

PR #48, opened 2017-01-16
For #38926, boostrap into project

jfboismenu:
Adds support for zero-config based projects.

### DIFF
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -355,25 +355,6 @@ class DesktopEngineSiteImplementation(object):
         """
         return self._current_login
 
-    def check_login_based(self, core_path):
-        """
-        Caches whether a pipeline configuration is login based or not.
-
-        :param core_path: Path to the core.
-        """
-        # Look inside the shotgun.yml file if there is a script user. If there isn't
-        # the core is login based and we have to refresh the credentials everytime
-        # we send a command to avoid password prompting in the background process.
-        shotgun_yml_path = os.path.join(core_path, "config", "core", "shotgun.yml")
-        with open(shotgun_yml_path, "r") as shotgun_yaml_file:
-            data = yaml.load(shotgun_yaml_file)
-            # If there are non null values set on both keys, we are not login based.
-            if data.get("api_script") and data.get("api_key"):
-                self._is_login_based = False
-            else:
-                self._is_login_based = True
-        self._engine.log_debug("login based: %s" % self._is_login_based)
-
     def get_current_user(self):
         """
         Returns the current login based user.
@@ -387,8 +368,7 @@ class DesktopEngineSiteImplementation(object):
         Refreshes the human user credentials, potentially prompting for a password, only is
         the desktop project engine is using login based authentication.
         """
-        if self._is_login_based:
-            self._user.refresh_credentials()
+        self._user.refresh_credentials()
 
 
 class KeyedDefaultDict(collections.defaultdict):
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -20,6 +20,7 @@ from tank.platform.qt import QtCore, QtGui
 
 import sgtk
 from sgtk.util import shotgun
+from sgtk.bootstrap import ToolkitManager
 from sgtk import util
 from sgtk.platform import constants
 from tank_vendor import shotgun_authentication as sg_auth
@@ -71,6 +72,18 @@ class DesktopWindow(SystrayWindow):
         self.__pipeline_configuration_separator = None
         self._settings_manager = settings.UserSettings(sgtk.platform.current_bundle())
 
+        engine = sgtk.platform.current_engine()
+
+        self._toolkit_manager = ToolkitManager(engine.get_current_user())
+        self._toolkit_manager.caching_policy = ToolkitManager.CACHE_FULL
+        self._toolkit_manager.plugin_id = os.environ.get(
+            "SGTK_DESKTOP_PROJECT_PLUGIN_ID", "config.basic"
+        )
+        self._toolkit_manager.base_configuration = os.environ.get(
+            "SGTK_DESKTOP_PROJECT_DESCRIPTOR_FALLBACK",
+            "sgtk:descriptor:app_store?name=tk-config-basic"
+        )
+
         # setup the window
         self.ui = desktop_window.Ui_DesktopWindow()
         self.ui.setupUi(self)
@@ -92,8 +105,6 @@ class DesktopWindow(SystrayWindow):
         self.ui.apps_button.setProperty("active", True)
         self.ui.apps_button.style().unpolish(self.ui.apps_button)
         self.ui.apps_button.style().polish(self.ui.apps_button)
-
-        engine = sgtk.platform.current_engine()
 
         connection = engine.get_current_user().create_sg_connection()
 
@@ -664,6 +675,8 @@ class DesktopWindow(SystrayWindow):
             engine = sgtk.platform.current_engine()
             engine.log_debug("launching app proxy for project: %s" % project)
 
+            # Phase 1: Get the UI pretty.
+
             # Make sure that not only the previous proxy is not running anymore
             # but that the UI has been cleared as well.
             engine = sgtk.platform.current_engine()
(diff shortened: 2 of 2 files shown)


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -355,25 +355,6 @@ def get_current_login(self):
         """
         return self._current_login
 
-    def check_login_based(self, core_path):
```
jfboismenu: This whole need to check if something is login based is a very complicated means to simply refresh the credentials if the user is login based.

python/tk_desktop/desktop_window.py
```diff
@@ -681,26 +696,24 @@ def __launch_app_proxy_for_project(self, project, pipeline_configuration_id=None
...
-                raise SystemError("Unsupported platform: %s" % sys.platform)
-
-            filters = [
-                ["project", "is", project],
-            ]
-
-            fields = [path_field, "users", "code"]
+            # Phase 2: Get information about the pipeline configuration.
 
+            # First get the legacy pipeline configurations, i.e. those using the *_path fields.
```
manneohrstrom: would be great to explain this a bit more - i think it will help future readers of this code. Going back to the diagrams, would be cool if we could communicate which cases are handled by bootstrap and which ones need to be handled by desktop for legacy purposes:

image

I don't think we should call these configs 'legacy' just yet. They are basically configs that are installed via tank setup_project or the install wizard - features that we haven't deprecated (and won't deprecate any time soon! - lots of clients depend on these workflows). So maybe call it the classic or installed workflow instead?

python/tk_desktop/desktop_window.py
```diff
@@ -681,26 +696,24 @@ def __launch_app_proxy_for_project(self, project, pipeline_configuration_id=None
...
             connection = engine.shotgun
             pipeline_configurations = connection.find(
                 "PipelineConfiguration",
-                filters,
-                fields=fields,
+                [["project", "is", project]],
+                fields=["mac_path", "windows_path", "linux_path", "users", "code"]
+            )
+            # We don't filter in the Shotgun query for the plugin ids because not every site these fields yet.
```
manneohrstrom: FYI heads up that the sg api handles non-existent fields this way:

```
>>> shotgun.find_one("Project", [], ["name", "field_that_doesnt_exist"])
{'type': 'Project', 'id': 63, 'name': 'Template Project'}
```
They are basically not returned in the return dataset.

(however you cannot use non-existent fields in filters!)

python/tk_desktop/desktop_window.py
```diff
@@ -681,26 +696,24 @@ def __launch_app_proxy_for_project(self, project, pipeline_configuration_id=None
...
+            # So if any pipeline configurations with a plugin id was returned, filter them it out.
+            pipeline_configurations = filter(
+                lambda pc: not("sg_plugin_ids" in pc or "plugin_ids" in pc), pipeline_configurations
+            )
+
+            # Gets the pipeline configurations that use plugin ids while respecting the users attribute.
```
manneohrstrom: what does "while respecting the users attribute" mean?

python/tk_desktop/desktop_window.py
```diff
@@ -681,26 +696,24 @@ def __launch_app_proxy_for_project(self, project, pipeline_configuration_id=None
...
+            pipeline_configurations.extend(
+                list(self._toolkit_manager.enumerate_pipeline_configurations(project))
```
manneohrstrom: see comment in core pr - i think we should KISS here and just return a list.

python/tk_desktop/desktop_window.py
```diff
@@ -715,27 +728,28 @@ def __launch_app_proxy_for_project(self, project, pipeline_configuration_id=None
             pipeline_configuration = None
             primary_pipeline_configuration = None
             for pc in pipeline_configurations:
+                # If we've stumbled upon the Primary.
                 if pc["code"] == constants.PRIMARY_PIPELINE_CONFIG_NAME:
                     primary_pipeline_configuration = pc
+                    # And there was no pipeline configuration saved from a last run.
```
manneohrstrom: nipicky: does `0` indicate that no pipeline config has been saved? Why not use `None`? That seems a little more standard to me!

manneohrstrom: could the code be made easier to read if we named the last accessed pipeline config id something explicit rather than just putting that value into `pipeline_configuration_id` (line 722)?

python/tk_desktop/desktop_window.py
```diff
@@ -715,27 +728,28 @@ def __launch_app_proxy_for_project(self, project, pipeline_configuration_id=None
...
             if pipeline_configuration is None:
-                if primary_pipeline_configuration is None:
-                    # Show the Setup Project widget
-                    self.setup_project_widget.project = project
-                    self.setup_project_widget.show()
-                    self.project_overlay.hide()
-                    return
-                else:
+                # ... but the primary exists, switch to that.
+                if primary_pipeline_configuration is not None:
```
manneohrstrom: with boostrap there could be as many as 3 primaries now:

image

Are we just displaying the "most important" of the three? or all of them?

how is this indicated in the UI? Would be great if we add some comments in the code to explain this particular case!

python/tk_desktop/desktop_window.py
```diff
@@ -756,27 +785,22 @@ def __launch_app_proxy_for_project(self, project, pipeline_configuration_id=None
...
-                                 "Configuration \"%s\" (id %d)." %
-                                 (current_platform,
-                                  pipeline_configuration["code"],
-                                  pipeline_configuration_id))
-
-                raise RuntimeError("The Toolkit configuration path has not\n"
-                                   "been set for your operating system.")
-
-            current_config_path = config_path
+            # FIXME: This logic should probably be built into core and be made public.
```
manneohrstrom: totally agree that we should add this to core. Now seems like as good a time as ever!

manneohrstrom:
one more comment: once we know the relevant core version, we should bump the min required version of core for the desktop engine.

python/tk_desktop/desktop_window.py
```diff
@@ -71,6 +72,20 @@ def __init__(self, parent=None):
         self.__pipeline_configuration_separator = None
         self._settings_manager = settings.UserSettings(sgtk.platform.current_bundle())
 
+        engine = sgtk.platform.current_engine()
+
+        self._toolkit_manager = ToolkitManager(engine.get_current_user())
+        self._toolkit_manager.caching_policy = ToolkitManager.CACHE_FULL
+        self._toolkit_manager.plugin_id = "config.basic"
```
jfboismenu: Should it? It's the configuration for the project, not the desktop. Surely you don't want a project specific config to use the same plugin id as the one for the desktop?

python/tk_desktop/desktop_window.py
jfboismenu: I've been updating the code everywhere calling it classic instead of installed because you could have a pc with hardcoded paths and a plugin id.
so classic means w/ path and w/o plugin id.

python/tk_desktop/desktop_window.py
jfboismenu: Exactly what my comment was about. Maybe I should clarify it! :)

python/tk_desktop/desktop_window.py
jfboismenu: Maybe we should create a separate ticket for this issue? Seems like a UX problem that should be resolved separately since it will involve discussions with more people.

python/tk_desktop/desktop_window.py
```diff
@@ -800,62 +800,13 @@ def report_progress(percentage, message):
...
-                    engine.log_error("No parent or interpreter found at '%s'."
-                                     % current_root_path)
-                    raise RuntimeError("The Toolkit configuration path points\n"
-                                       "to an invalid configuration.")
-
-                # Read the path to the parent configuration
-                with open(parent_config_file, "r") as f:
-                    current_root_path = f.read().strip()
+            try:
+                path_to_python = sgtk.get_python_interpreter_for_config(config_path)
```
manneohrstrom: nitpick and i also haven't looked at the core CR yet - but wondering whether this should be `stgk.util` rather than just `sgtk`?

python/tk_desktop/desktop_window.py
```diff
@@ -681,128 +714,99 @@ def __launch_app_proxy_for_project(self, project, pipeline_configuration_id=None
...
-            ]
+            # Step 1: First get the legacy pipeline configurations, i.e. those using the *_path fields.
+            connection = engine.shotgun
+            pipeline_configurations = self._get_toolkit_classic_pipeline_configurations(connection, project)
 
-            fields = [path_field, "users", "code"]
+            toolkit_manager = ToolkitManager(engine.get_current_user())
+            # We need to cache all environments because we don't know which one the user will require.
+            toolkit_manager.caching_policy = ToolkitManager.CACHE_FULL
+            toolkit_manager.plugin_id = "config.desktop"
```
manneohrstrom: would it make sense to turn plugin id and base config params into constants?

python/tk_desktop/desktop_window.py
```diff
@@ -681,128 +714,99 @@ def __launch_app_proxy_for_project(self, project, pipeline_configuration_id=None
...
-                                       "to an invalid configuration.")
-
-                # Read the path to the parent configuration
-                with open(parent_config_file, "r") as f:
-                    current_config_path = f.read().strip()
+                # We did have something in Shotgun that was selected, let's pick that for bootstrapping.
+                toolkit_manager.pipeline_configuration = most_recent_pipeline_configuration["id"]
+
+            # Make sure the config is downloaded and the bundles cached.
+            config_path = toolkit_manager.prepare_engine(None, project)
```
manneohrstrom: `toolkit_manager.prepare_engine("tk-desktop", project)`?

PR #50, opened 2017-01-27
For #38927, User menu item to launch classic Project Setup Wizard

akhronek:
Adds an 'Advanced project setup...' action to the User menu for projects that bootstrap Toolkit using the tk_config_basic configuration.  This menu item is not available on the Projects page or for projects that have an existing pipeline configuration in Shotgun. Images are posted in the title ticket.

Also implements a slot to add help popup data for the Setup Project Wizard, wasn't sure what to put there at this time.


python/tk_desktop/desktop_window.py
```diff
@@ -368,6 +375,27 @@ def handle_project_refresh_action(self):
...
+        setup a project. If not, pop up an error dialog informing
+        them of the problem.
+        """
+        self.setup_project_widget.project = self.current_project
+
+        # The first time a user selects the Advanced project setup
+        # menu item, display the Setup Project help popup to provide
+        # more information about this feature.
+        wizard_setting = "advanced_project_setup_launched"
+        wizard_help_shown = self._load_setting(wizard_setting, False, False)
```
jfboismenu: Could you pass named parameters for he booleans? It would help the reader understand what those values are without having to scroll back to the documentation of the method.

python/tk_desktop/desktop_window.py
```diff
@@ -368,6 +375,27 @@ def handle_project_refresh_action(self):
...
+        if not wizard_help_shown:
+            self._save_setting(wizard_setting, True, False)
```
jfboismenu: Same comment here about booleans

python/tk_desktop/setup_project.py
```diff
@@ -62,10 +69,57 @@ def do_setup(self):
                                        "and delete all pipeline configurations for your project.")
             ret = error_dialog.exec_()
```
jfboismenu: ret is assigned at t couple of places but I don't see it being returned anywhere? Dead code? Might be worth removing it.

python/tk_desktop/setup_project.py
```diff
@@ -62,10 +69,57 @@ def do_setup(self):
...
+            sg_project = engine.shotgun.find_one(
+                "Project", [["id", "is", self.project["id"]]], ["tank_name"]
+            )
+            engine.shotgun.update(
+                "Project", self.project["id"], {"tank_name": "foobar"}
+            )
+            engine.shotgun.update(
+                "Project", self.project["id"], {"tank_name": sg_project["tank_name"]}
+            )
+        except Exception, e:
```
jfboismenu: Might be worth catching an shotgun_api3.Fault instead to announce in advance the intent of the catch.

python/tk_desktop/desktop_window.py
```diff
@@ -818,10 +848,16 @@ def report_progress(percentage, message):
             # comes from.
```
jfboismenu: I think my comment induced you in error here. My comment should have read

`If no pipeline configuration is in the user settings, we will let the bootstrap pick the right pipeline configuration for the first launch.`

See next comment.

python/tk_desktop/desktop_window.py
```diff
@@ -818,10 +848,16 @@ def report_progress(percentage, message):
...
             if most_recent_pipeline_configuration is None:
                 toolkit_manager.pipeline_configuration = None
+
+                # Enable user menu item to launch classic Project Setup wizard
```
jfboismenu: Here the comment should have been
`We've loaded this project before and saved its pipeline configuration id, so reload the same old one.`

jfboismenu: In general, pipeline configurations not using plugin ids have paths associated to them (Toolkit classic), so the configurations returned from `_get_toolkit_classic_pipeline_configurations` could has a boolean called classic. Then we could add the flag for entries coming back from `toolkit_manager.get_pipeline_configurations` before extending the list `pipeline_configurations`. Then test if the selected pipeline configuration is `not None` and doesn't have the `classic` flag. Those would be the pipeline configurations using a plugin id.

We do have an edge case where the pipeline configuration might have a plugin id AND a path for all platforms. In that case we wouldn't want to show the menu entry, but the problem here is that we are not exposing that information in get_pipeline_configurations. I think this needs to be discussed with @manneohrstrom  or over Dailies.

python/tk_desktop/setup_project.py
```diff
@@ -60,12 +68,59 @@ def do_setup(self):
...
         finally:
             if is_on_top: 
                 self._set_top_window_on_top(True)
 
+    def show_help_popup(self):
+        """
+        Display a help screen
+        """
+        # Not sure what to put here.
+        pass
```
akhronek: Any ideas for what should be implemented in this spot?  The launcher help popup looks really great.  Should we try to emulate something like that here? Or is a nicely formatted link to the existing docs sufficient?

python/tk_desktop/desktop_window.py
```diff
@@ -810,16 +844,26 @@ def report_progress(percentage, message):
...
 
-            # If no pipeline configuration is in Shotgun, we'll let the bootstrap decide where the config
-            # comes from.
+            # If no pipeline configuration is in the user settings, we will let the bootstrap
+            # pick the right pipeline configuration for the first launch.
             if most_recent_pipeline_configuration is None:
                 toolkit_manager.pipeline_configuration = None
             else:
-                # We did have something in Shotgun that was selected, let's pick that for bootstrapping.
+                # We've loded this project before and saved its pipeline configuation id, so
```
jfboismenu: loaded

python/tk_desktop/setup_project.py
jfboismenu: I think this is more of a question for these guys: @manneohrstrom @robblau

jfboismenu: After discussing with Rob, let's put a link in there for now and then worry about having a tk-multi-loader-like help pop-up in 0-config part deux.

PR #51, opened 2017-01-27
For #39292, Support transparent Software entity icons

akhronek:
Updates required to retrieve original source images uploaded for thumbnails. The primary reason for this update was to get the icons associated with Software entities to display with transparency in the NoAppsInstalledOverlay widget. The implementation attempts to handle older versions of tk-core and tk-framework-shotgunutils that may not have this feature implemented.

### DIFF
--- a/python/tk_desktop/no_apps_installed_overlay.py
+++ b/python/tk_desktop/no_apps_installed_overlay.py
@@ -108,7 +108,8 @@ class NoAppsInstalledOverlay(QtGui.QWidget):
                 engine.logger.warning(
                     "ShotgunDataRetriever is unable to download thumbnail source file. "
                     "Attempting to download thumbnail instead. This issue may be "
-                    "resolved by updating local installations of tk-framework-shotgunutils."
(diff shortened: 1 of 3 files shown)


python/tk_desktop/no_apps_installed_overlay.py
```diff
@@ -94,14 +95,30 @@ def _load_software_icons(self, engine, project=None):
...
+                # transparency and alpha values.
+                sg_icon = shotgun_data.ShotgunDataRetriever.download_thumbnail_source(
+                    sg_softwares[i]["type"], sg_softwares[i]["id"], engine
+                )
+
+            except (TankError, AttributeError):
+                # An old version of tk-framework-shotgunutils or tk-core is 
+                # likely in use. Try ShotgunDataRetriever.download_thumbnail() instead.
+                engine.logger.warning(
+                    "ShotgunDataRetriever is unable to download thumbnail source file. "
```
jfboismenu: Maybe the message could say something about what the end result is, i.e. having icons with no transparency. This would help the user understand why upgrading is important.

PR #53, opened 2017-02-01
For #39698, desktop installer with bundle cache

jfboismenu:
Pushes the bundle cache from the site config into the project's.


python/tk_desktop/desktop_window.py
```diff
@@ -762,13 +762,9 @@ def __launch_app_proxy_for_project(self, project, pipeline_configuration_id=None
             toolkit_manager.caching_policy = ToolkitManager.CACHE_FULL
             toolkit_manager.plugin_id = "config.desktop"
             toolkit_manager.base_configuration = "sgtk:descriptor:app_store?name=tk-config-basic"
-
-            # FIXME: This needs to be replaced when we implement proper progress reporting during
```
jfboismenu: Old code that should have been deleted and was overriden later down

PR #54, opened 2017-02-01
Addendum for #39721, add progress indicator

jfboismenu:
Makes sure the overlay is reset between uses.

### DIFF
--- a/python/tk_desktop/loading_project_widget.py
+++ b/python/tk_desktop/loading_project_widget.py
@@ -74,6 +74,9 @@ class LoadingProjectWidget(QtGui.QWidget):
         """
         Starts the progress reporting.
         """
+        # Reset the message
+        self._message = None
+        self._show_hide_widgets(show=True)
         self._ui.shotgun_spinning_widget.start_progress()
         # Hide details
         self._ui.progress_output.hide()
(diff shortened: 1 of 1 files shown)


python/tk_desktop/loading_project_widget.py
```diff
@@ -107,14 +110,22 @@ def show_error_message(self, msg):
...
-        self._ui.shotgun_spinning_widget.hide()
-        self._ui.bottom.hide()
-        self._ui.progress_output.hide()
+        self._show_hide_widgets(show=False)
 
         self.setVisible(True)
         self._message = msg
         self.repaint()
 
+    def _show_hide_widgets(self, show=False):
```
josh-t: I think this should just be called `_show_widgets()`. The boolean argument implies you're either showing or not showing I think. The calls above become:

`self._show_widgets(True)`

and 

`self._show_widgets(False)`

python/tk_desktop/loading_project_widget.py
```diff
@@ -107,14 +110,22 @@ def show_error_message(self, msg):
...
+        """
+        Shows or hides the widgets in the UI.
+
+        :param show: If True, shows all the widgets in the UI.
```
josh-t: I'd amend this just to make it clear that a `False` argument hides all the widgets in the UI.

PR #56, opened 2017-02-04
For #38926, boostrap into project

jfboismenu:
Since more than two pipelines can have the same name, we suffix them with their id when that happens. Does also various bugfixes.

Here's what it looks like in action.
screen shot 2017-02-03 at 23 13 56

I don't really like using the id when two things are named the same, but at least it is consistent with other areas in Shotgun like the Shotgun Model.  Ideally we should add a tooltip with the string of the descriptor in the menu or put the descriptor itself inside the parantheses.

### DIFF
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -71,6 +71,8 @@ class DesktopWindow(SystrayWindow):
 
     ORGANIZATION = "Shotgun Software"
     APPLICATION = "tk-desktop"
+    _BOOTSTRAP_END_RATIO = 0.9
+    _LAUNCHING_PYTHON_RATIO = 0.95
 
     def __init__(self, parent=None):
         SystrayWindow.__init__(self, parent)
@@ -293,7 +295,7 @@ class DesktopWindow(SystrayWindow):
 
                 win32api.SetDllDirectory(None)
             except StandardError:
-                log.warning('Could not push DllDirectory under Windows.')
+                log.warning("Could not push DllDirectory under Windows.")
 
     def _pop_dll_state(self):
         '''
@@ -304,7 +306,7 @@ class DesktopWindow(SystrayWindow):
                 import win32api
                 win32api.SetDllDirectory(self._previous_dll_directory)
             except StandardError:
-                log.warning('Could not restore DllDirectory under Windows.')
+                log.warning("Could not restore DllDirectory under Windows.")
 
     ########################################################################################
     # Event handlers and slots
@@ -682,12 +684,12 @@ class DesktopWindow(SystrayWindow):
 
             # If this pipeline is the one that was selected, mark it in the
             # menu and update the configuration name widget.
-            if selected["id"] == pc["id"]:
+            if selected and selected["id"] == pc["id"]:
                 action.setChecked(True)
                 self.ui.configuration_name.setText(unique_pc_name)
 
                 # If we haven't picked a primary, show the sandbox header.
-                if (pc["name"] != constants.PRIMARY_PIPELINE_CONFIG_NAME):
+                if not self._is_primary_pc(pc):
                     self.ui.configuration_frame.show()
 
     def __set_project_just_accessed(self, project):
@@ -802,7 +804,7 @@ class DesktopWindow(SystrayWindow):
                 return False
 
             if not ShotgunPath.from_shotgun_dict(pc):
-                log.warning("Skipping Toolkit Classic pipeline configuration id %s without any path set." % pc["id"])
+                log.warning("Skipping Toolkit Classic pipeline configuration '%s' without any path set." % pc["id"])
                 return False
 
             # If there are no users assigned, this pipeline is accessible from for everyone.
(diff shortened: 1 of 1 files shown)


manneohrstrom:
Regarding the display and the screenshot above - we shouldn't display multiple primaries in the dropdown - this feels extremely confusing for the user. Here's a breakdown of how we should present things in order to be consistent with our current behaviour - hope this helps clarify things. The big change here from what we have been doing previously is that the bootstrap allows for site level dev sandboxes which apply to all projects. In the design image below, i suggest a helpful *(site)* suffix for such entries to help indicate that they are site level configs:

overrides

jfboismenu:
I would argue that we should still do something about the multiple primaries issue. As it stands if the user has somehow configured multiple primaries for a single project we should at the very least show a warning in the console (and based on our discussion today on auto detection, pick the one with the lowest id).

python/tk_desktop/desktop_window.py
```diff
@@ -738,31 +785,78 @@ def _get_toolkit_classic_pipeline_configurations(self, connection, project):
...
+
+            # Search for ourselves.
+            for user in pc["users"]:
+                # We've found ourselves, awesome!
+                if self._current_user_id == user["id"]:
+                    return True
+
+            return False
+
+        # FIXME: We need to discuss what we do with multiple primaries with Toolkit classic. In zero config we keep
```
jfboismenu: Let's discuss this on Thursday @manneohrstrom , I've already added this to the catch-up document.

jfboismenu:
Updated so that
- Toolkit classic primary config shadows zero-config config, so only one primary is available
- all Toolkit classic and zero-config sandboxes are available
screen shot 2017-02-14 at 11 47 07

python/tk_desktop/desktop_window.py
```diff
@@ -286,8 +293,7 @@ def _push_dll_state(self):
 
                 win32api.SetDllDirectory(None)
             except StandardError:
-                engine = sgtk.platform.current_engine()
-                engine.log_warning('Could not push DllDirectory under Windows.')
+                log.warning('Could not push DllDirectory under Windows.')
```
manneohrstrom: double quotes pls :)

python/tk_desktop/desktop_window.py
```diff
@@ -298,8 +304,7 @@ def _pop_dll_state(self):
                 import win32api
                 win32api.SetDllDirectory(self._previous_dll_directory)
             except StandardError:
-                engine = sgtk.platform.current_engine()
-                engine.log_warning('Could not restore DllDirectory under Windows.')
+                log.warning('Could not restore DllDirectory under Windows.')
```
manneohrstrom: double quotes pls :)

python/tk_desktop/desktop_window.py
```diff
@@ -594,54 +599,96 @@ def show_update_project_config(self):
...
+            # If this pipeline is the one that was selected, mark it in the
+            # menu and update the configuration name widget.
             if selected["id"] == pc["id"]:
-                self.ui.configuration_frame.show()
                 action.setChecked(True)
-                self.ui.configuration_name.setText(pc["name"])
+                self.ui.configuration_name.setText(unique_pc_name)
+
+                # If we haven't picked a primary, show the sandbox header.
+                if (pc["name"] != constants.PRIMARY_PIPELINE_CONFIG_NAME):
```
manneohrstrom: don't need the brackets

python/tk_desktop/desktop_window.py
```diff
@@ -738,31 +785,78 @@ def _get_toolkit_classic_pipeline_configurations(self, connection, project):
...
+            fields=["mac_path", "windows_path", "linux_path", "users", "code", "sg_plugin_ids", "plugin_ids", "project"]
         )
 
-        for pc in pipeline_configurations:
+        def is_pipeline_usable(pc):
+            """
+            Ensures that a pipeline is not zero-config based and that it is accessible for the current user.
+            """
+            # If there is a plugin id, can't use this.
+            if pc.get("sg_plugin_ids") or pc.get("plugin_ids"):
```
manneohrstrom: could this be filtered out in the sg query?

python/tk_desktop/desktop_window.py
```diff
@@ -738,31 +785,78 @@ def _get_toolkit_classic_pipeline_configurations(self, connection, project):
...
+                return False
+
+            if not ShotgunPath.from_shotgun_dict(pc):
```
manneohrstrom: nice!

python/tk_desktop/desktop_window.py
manneohrstrom: cool!

classic takes precedence, so maybe it can just check if there is a classic config - in that case, that is used, if not we fall back on bootstrap logic. Apologies if i am over simplifying :)

python/tk_desktop/desktop_window.py
```diff
@@ -738,31 +785,78 @@ def _get_toolkit_classic_pipeline_configurations(self, connection, project):
...
+
+        # Massage the dictionaries keys so it uses the same key names of the zero config pipelines.
+        for pc in accessible_pcs:
             pc["name"] = pc["code"]
             del pc["code"]
 
-        return pipeline_configurations
+        return accessible_pcs
+
+    def _is_primary_pc(self, pc):
```
manneohrstrom: with all these helper methods it feels like we should almost split out this list of pipeline configs, with all its various utility methods and business logic, into its own class. for modularity. what do you think?

python/tk_desktop/desktop_window.py
```diff
@@ -738,31 +785,78 @@ def _get_toolkit_classic_pipeline_configurations(self, connection, project):
...
+        """
+        Tests if a pipeline configuration is a primary.
+
+        :param pc: Pipeline configuration entity with key ``code``.
+
+        :returns: True if the pipeline configuration is a primary, else otherwise.
+        """
+        return pc["name"] == constants.PRIMARY_PIPELINE_CONFIG_NAME
+
+    def _merge_pipeline_configuration_lists(self, classic_pipelines, zero_config_pipelines):
```
manneohrstrom: here we are using `pipeline` again

manneohrstrom: how about we say `bootstrap_pipeline_configs` instead of zero config. Zero config for me is just one possible configuration, but they all come from the bootstrap API.

python/tk_desktop/desktop_window.py
```diff
@@ -890,6 +947,65 @@ def report_progress(percentage, message):
         self._sync_thread.sync_success.connect(self._sync_success)
         self._sync_thread.start()
 
+    def _pick_pipeline(self, pipeline_configurations, requested_pipeline_id, project):
```
manneohrstrom: pipeline

python/tk_desktop/desktop_window.py
jfboismenu: We can't filters on fields that may not exist on the schema. We have the same problem in the core's resolver.

jfboismenu:
I totally agree that we should break out lots of code from that file. That was my intent when I suggested that we should schedule a ticket for writing unit tests for the engine. I would rather postpone those changes for when I actually write the tests so I'll have a better idea of what I need to test and how I need to refactor everything. How does that sound?

python/tk_desktop/desktop_window.py
jfboismenu: I'm not sure what shape that would take. Do you mean something in core? I'd LOVE to have wrappers around Shotgun entities that would give you some syntaxic sugar on top of the entity dictionaries. Ideally those wrappers would be returned by the Shotgun Api and not something you manually wrap. I suppose that's not happening anytime soon tough, but a man can dream, right?

jfboismenu: Maybe we could have a spike on this where we take into account our usage of the Shotgun api and devise a plan around having utility methods, wrappers or something even more cooler that wraps the Shotgun API?

jfboismenu: That was actually one of my hackathon ideas: a Shotgun API that returns Python objects whose interface is inferred from the schema.

python/tk_desktop/desktop_window.py
manneohrstrom: ah yes. bummer!

python/tk_desktop/desktop_window.py
manneohrstrom: hehe, i was thinking something much simpler. My question was basically - "maybe now that we are adding even more stuff to this 1000+ lines file, maybe we can break it into multiple files? If those files contain classes, methods etc" - just basic housekeeping. That's all i was suggesting :) Nothing fancy :)

manneohrstrom: There is a whole bunch of pipeline configuration related logic and methods in this file currently - maybe we could somehow move that out into a separate file - still in desktop, just a reshuffle to make it easier to maintain the code basically?

python/tk_desktop/desktop_window.py
```diff
@@ -58,12 +63,16 @@
 
 ShotgunModel = shotgun_model.ShotgunModel
 
+log = get_logger(__name__)
```
manneohrstrom: nice!

PR #57, opened 2017-02-08
#40549, Support 'group' and 'group_default' command properties

akhronek:
If no collapse rules have been determined for a command, check for input `group` and `group_default` properties. If present, these will be treated the same as collapse rules. The current implementation does not actually support the `group_default` property, as this is determined at a much lower level by the project_commands_model.

### DIFF
--- a/python/tk_desktop/desktop_engine_project_implementation.py
+++ b/python/tk_desktop/desktop_engine_project_implementation.py
@@ -119,7 +119,7 @@ class DesktopEngineProjectImplementation(object):
             self.__callback_map[("__commands", name)] = command_info["callback"]
             # pull out needed values since this needs to be pickleable
             gui_properties = {}
-            for prop in ["type", "icon", "title", "description"]:
+            for prop in ["type", "icon", "title", "description", "group", "group_default"]:
                 if prop in command_info["properties"]:
                     gui_properties[prop] = command_info["properties"][prop]
             # evaluate groups on the app proxy side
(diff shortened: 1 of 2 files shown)


python/tk_desktop/desktop_engine_project_implementation.py
```diff
@@ -119,7 +119,7 @@ def _register_commands(self):
             self.__callback_map[("__commands", name)] = command_info["callback"]
             # pull out needed values since this needs to be pickleable
             gui_properties = {}
-            for prop in ["type", "icon", "title", "description"]:
+            for prop in ["type", "icon", "title", "description", "group", "group_default"]:
```
manneohrstrom: Would be great if we could add these options to the docs here: http://developer.shotgunsoftware.com/tk-core/platform.html#sgtk.platform.Engine.register_command

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -146,6 +152,8 @@ def action_triggered():
             # the display name of the command
             menu_name = None
             button_name = title
+            found_collapse_match = False
```
manneohrstrom: it would be great with some comments here (with examples) explaining the collapsing and the grouping, how they relate and what takes precedence. This code was already kind of hard to read and now we just added more complexity, so a bit of comments to help out future code readers i think would be really helpful!

PR #58, opened 2017-02-23
40549 improved grouping support

akhronek:
Implementation to support the `group_default` property value passed into the `engine.register_command()` method.  Also fixes an issue where icons were not being loaded for non-default commands in the RECENTS group.

### DIFF
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -109,9 +109,15 @@ class DesktopEngineSiteImplementation(object):
         command_type = properties.get("type")
         command_icon = properties.get("icon")
         command_tooltip = properties.get("description")
+        command_group = properties.get("group")
+        command_is_group_default = properties.get("group_default")
+        if not command_group:
+            # If no group has been specified,then this command is automatically
+            # the group default.
+            command_is_group_default = True
 
         icon = None
-        if command_icon is not None:
+        if command_icon is not None and command_is_group_default:
             if os.path.exists(command_icon):
                 icon = QtGui.QIcon(command_icon)
             else:
@@ -146,6 +152,8 @@ class DesktopEngineSiteImplementation(object):
             # the display name of the command
             menu_name = None
             button_name = title
+            found_collapse_match = False
+
             for collapse_rule in self._collapse_rules:
                 template = DisplayNameTemplate(collapse_rule["match"])
                 match = template.match(title)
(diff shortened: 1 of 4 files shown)


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -178,7 +173,7 @@ def action_triggered():
                 menu_name = title
 
             self.desktop_window.add_project_command(
-                name, button_name, menu_name, icon, command_tooltip, groups
+                name, button_name, menu_name, icon, command_tooltip, groups, command_is_menu_default
```
manneohrstrom: it's a pretty long line now. i suggest splitting it so we have a single param on each line!

python/tk_desktop/desktop_window.py
```diff
@@ -524,10 +524,10 @@ def add_to_project_menu(self, action):
         self.project_menu.insertAction(self.__pipeline_configuration_separator, action)
 
     def add_project_command(
-            self, name, button_name, menu_name, icon, command_tooltip, groups
-    ):
+            self, name, button_name, menu_name, icon, command_tooltip, groups, is_menu_default
+        ):
         self._project_command_model.add_command(
-            name, button_name, menu_name, icon, command_tooltip, groups
+            name, button_name, menu_name, icon, command_tooltip, groups, is_menu_default
```
manneohrstrom: please add doc strings whenever you touch a method which doesn't have docstrings defined :)

python/tk_desktop/project_commands_model.py
```diff
@@ -139,7 +140,7 @@ def set_project(self, project, groups, show_recents=True):
             self.set_group_rank(self.RECENT_GROUP_NAME, 0)
             self.__load_recents()
 
-    def add_command(self, name, button_name, menu_name, icon, command_tooltip, groups):
+    def add_command(self, name, button_name, menu_name, icon, command_tooltip, groups, is_menu_default=False):
```
manneohrstrom: doc string

python/tk_desktop/project_commands_model.py
```diff
@@ -212,8 +232,14 @@ def child_cmp(left, right):
             if util.is_version_older(left_version, right_version):
                 return 1
             return 0
-        children.sort(cmp=child_cmp)
-        return children
+
+        # No need to sort lists if they are empty.
+        if other_children:
```
manneohrstrom: this feels like premature optimization to me. it's even possible that it's actually faster without the check :) I suggest removing it.

python/tk_desktop/project_commands_model.py
```diff
@@ -186,22 +187,41 @@ def add_command(self, name, button_name, menu_name, icon, command_tooltip, group
...
+
+        :param QtGui.QStandardItem item: Input project button to sort child menu items for.
+        :returns: List of sorted menu items
+        """
+        # Theoretically, there should only be one "default" child in the list of item
+        # children, but since this is a human configurable value, there may 'accidentally'
+        # be more. Rather than raising an error, attempt to handle this case gracefully
+        # by keeping track of default and non-default children separately. Sort each
+        # list in reverse version number order, then add them together to construct
+        # the list of ordered item children to return.
```
manneohrstrom: an example here would help a great deal. also, why not move the comment up in the doc string?

PR #59, opened 2017-03-13
For #36439, site and project desktop use same credentials

jfboismenu:
Passes down the credentials from the site desktop to the project desktop with an environment variable. The ticket originally suggested using the `TANK_CONTEXT` environment variable, but using a separate environment was used instead in order to both reduce the number of changes required to make this fix work as well as ensuring there wouldn't be some weird side-effects in we ended up not using the authenticated user from the context. This would have happened on projects that used a script to authenticate.

### DIFF
--- a/python/utils/bootstrap_utilities.py
+++ b/python/utils/bootstrap_utilities.py
@@ -35,12 +35,12 @@ def start_engine(data):
         current_user = ShotgunAuthenticator(sgtk.util.CoreDefaultsManager()).get_default_user()
 
         # If we found no user using the authenticator, we need to use the credentials that
-        # came through the data file.
+        # came through the environment variable.
         # Also, if the credentials are user-based, we need to disregard what we got and use
-        # the credentials from the data file. This is required to solve any issues
+        # the credentials from the environment variable. This is required to solve any issues
         # arising from the changes to the session cache changing place in core 0.18.
         if not current_user or current_user.login:
-            current_user = deserialize_user(data["current_user"])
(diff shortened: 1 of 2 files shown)


python/utils/bootstrap_utilities.py
```diff
@@ -85,6 +99,7 @@ def start_app(engine):
         engine.msg_server.join()
         return 0
 
+
```
jfboismenu: PEP8 :)

python/utils/bootstrap_utilities.py
```diff
@@ -31,9 +31,23 @@ def start_engine(data):
     # a current user, we have to set the authenticated user.
     if hasattr(sgtk, "set_authenticated_user"):
         # Retrieve the currently authenticated user for this process.
-        from tank_vendor.shotgun_authentication import ShotgunAuthenticator
-        user = ShotgunAuthenticator(sgtk.util.CoreDefaultsManager()).get_default_user()
-        sgtk.set_authenticated_user(user)
+        from tank_vendor.shotgun_authentication import ShotgunAuthenticator, deserialize_user
+        current_user = ShotgunAuthenticator(sgtk.util.CoreDefaultsManager()).get_default_user()
+
+        # If we found no user using the authenticator, we need to use the credentials that
```
manneohrstrom: Why is `SHOTGUN_DESKTOP_CURRENT_USER` a fallback? is it a performance thing? Shouldn't we always look for this and pick it up and use it? like a contract between the two RPC desktop processes? It might be worth adding a comment to clarify why we need to pick up the default user!

PR #68, opened 2017-05-04
For #42178. websocket callbacks

jfboismenu:
The Shotgun Desktop will now offer to restart when the user and or site making a request through websocket is different than the one running the Shotgun Desktop.
screen shot 2017-05-04 at 14 30 53
We're popping the browser into focus, both in pinned and unpinned mode so the user knows immediately what's up.

> Note that this branch builds on top of 41850_user_updates_banner. Once that first branch is merged we'll merge this one.


python/tk_desktop/browser_integration_user_switch_dialog.py
```diff
@@ -0,0 +1,63 @@
...
+
+        :param int state: State of the checkbox.
+        """
+        self.ui.restart_button.setEnabled(state != QtCore.Qt.Checked)
+
+    def _on_restart_clicked(self):
+        """
+        Dismisses the dialog with the the Restart code.
+        """
+        self.done(self.Restart)
```
thebeeland: [nitpick] At first glance it confused me seeing "Restart" here. Any reason the class variables aren't all caps?

python/tk_desktop/desktop_window.py
```diff
@@ -17,6 +17,7 @@
 import cPickle as pickle
 import pprint
 import itertools
+import urlparse
 from collections import OrderedDict
```
thebeeland: OrderedDict is 2.7+, isn't it? I guess since Desktop is bundled with 2.7 this is fine. Just mentioning it in case there's ever the possibility we won't be in 2.7.

python/tk_desktop/desktop_window.py
```diff
@@ -513,25 +524,128 @@ def on_project_commands_finished(self):
...
         engine = sgtk.platform.current_engine()
-
         try:
             sg_auth.ShotgunAuthenticator().clear_default_user()
             if engine.uses_legacy_authentication():
                 engine.create_legacy_login_instance().logout()
+            return True
         except Exception:
             # if logout raises an exception, just log and don't crash
             log.exception("Error logging out.")
```
thebeeland: [minor] Maybe log the exception message?

python/tk_desktop/desktop_window.py
```diff
@@ -513,25 +524,128 @@ def on_project_commands_finished(self):
...
+    def _on_different_user(self, site, user_id):
+        """
+        Invoked when a request coming from a different site and/or user comes through.
+
+        :param str site: URL of the site making the request.
+        :param int user_id: User id of the HumanUser attempting the request.
+        """
+        # Makes sure that if the user is browsing multiples pages before coming back to the Desktop,
+        # only the first request will generate a pop-up. Note that if requests comes from different users and/or sites,
+        # only the first one will be acknoledged. This is to avoid having multiple modal dialogs popping up.
```
thebeeland: "acknowledged"

python/tk_desktop/browser_integration_user_switch_dialog.py
jfboismenu: Oops!

python/tk_desktop/desktop_window.py
jfboismenu: Yeah, desktop is 2.7 only, it runs with the builtin Python, so there's no worry there.

python/tk_desktop/desktop_window.py
jfboismenu: `log.exception` logs the exception and the stack by definition.

```
import logging
try:
    raise Exception("This is bad!!!")
except Exception:
    logging.exception("Unexpected")
```

yields

```
ERROR:root:Unexpected
  File "<stdin>", line 2, in <module>
Exception: This is bad!!!
```

PR #70, opened 2017-05-16
For #42441, websocket back compat

jfboismenu:
Ensures that we don't start a websocket server a second time if the desktop startup already started its own server.

I also shortened a URL which had unnecessary bits in it.

### DIFF
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -265,11 +265,10 @@ class DesktopEngineSiteImplementation(object):
         f.close()
         app.setStyleSheet(css)
 
-        # If the desktop startup is not running a websocket server, then start the builtin one.
-        # The latest version of the desktop startup does not run the server anymore, but the
-        # original one did and if the desktop engine gets updated but not the startup then we are
-        # forced to let the other one run.
-        if "server" not in kwargs:
+        # Sadly, we can't tear down the previous server and restart it. Attempting to tear_down() and
+        # instantiate a new server will raise an error.ReactorNotRestartable exception. So we'll start
+        # our websocket integration only if there is no server running from the desktop startup.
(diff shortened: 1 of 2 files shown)


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -266,13 +265,18 @@ def run(self, splash, version, **kwargs):
...
-        # styled after the Shotgun Desktop's visual-style.
-        splash.set_message("Initializing browser integration.")
-        try:
-            desktop_server_framework.launch_desktop_server(self._user.host, self._current_login["id"])
-        except Exception:
-            logger.exception("Unexpected error while trying to launch the browser integration:")
+        # Sadly, we can't tear down the previous server and restart it. Attempting to tear_down() and
+        # instantiate a new server will raise an error.ReactorNotRestartable exception. So we'll start
+        # our websocket integration only if there is no server running from the desktop startup.
+        if kwargs.get("server") is None:
```
manneohrstrom: the comment doesn't mention why `server` is significant. If you could please elaborate that would be great!

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -266,13 +265,18 @@ def run(self, splash, version, **kwargs):
...
+            # Initialize all of this after the style-sheet has been applied so any prompt are also
+            # styled after the Shotgun Desktop's visual-style.
+            splash.set_message("Initializing browser integration.")
+            try:
+                desktop_server_framework = sgtk.platform.get_framework("tk-framework-desktopserver")
+                desktop_server_framework.launch_desktop_server(self._user.host, self._current_login["id"])
+            except Exception:
+                logger.exception("Unexpected error while trying to launch the browser integration:")
```
manneohrstrom: should we have an else here? perhaps print a debug message?

PR #72, opened 2017-05-24
For #41833 and #42258, desktop fixes

jfboismenu:
Fixed logging and recent item names.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -72,6 +72,13 @@ class DesktopEngine(Engine):
         if hasattr(self.__impl, "destroy_engine"):
             self.__impl.destroy_engine()
 
+    def _emit_log_message(self, handler, record):
+        """
+        Called when a message needs to be logged.
+        """
+        if hasattr(self.__impl, "_emit_log_message"):
+            self.__impl._emit_log_message(handler, record)
+
     ############################################################################
     # Dispatch to our implementation
     def __getattr__(self, attr):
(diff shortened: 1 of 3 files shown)


engine.py
```diff
@@ -72,6 +72,13 @@ def destroy_engine(self):
         if hasattr(self.__impl, "destroy_engine"):
             self.__impl.destroy_engine()
 
+    def _emit_log_message(self, handler, record):
```
jfboismenu: The desktop engine is broken in two halves. The part in engine.py, which is the base implementation common to the Shotgun Desktop and its background process and the `desktop_engine_{site,project}_implementation.py` file. The __impl is the implementation coming from that second file. Desktop engine's custom handlers are implemented in those files and when the core calls these methods they are forwarded to the appropriate location.

python/tk_desktop/desktop_engine_project_implementation.py
```diff
@@ -328,7 +328,7 @@ def _get_setting(self, setting_name, default_value=None):
         """
         return self._engine.get_setting(setting_name, default_value)
 
-    def _emit_log_message(self, level, msg, *args):
+    def _emit_log_message(self, handle, record):
```
jfboismenu: Don't know what I was thinking when I wrote this originally. It made no sense. This is NOT the signature of `sgtk.platform.Engine._emit_log_message`

python/tk_desktop/project_commands_widget.py
```diff
@@ -162,7 +162,7 @@ def _text_for_item(self, item):
         if menu_name is None:
             return button_name
         else:
-            return "%s\n%s" % (button_name, menu_name)
+            return menu_name
```
jfboismenu:
```
Maya
Maya 2017
```

now becomes
`Maya 2017`

I've also validated that this works with classic configs and apps like the publisher.

PR #76, opened 2017-06-07
For #43159, chrome 59 fix

jfboismenu:
Chrome 59 on Mac breaks browser integration because our certificate doesn't report the correct certificate version. Version 3 of the certificate is the one that accepts X5009 extensions.

We're adding a new `Advanced` submenu in the user menu, which now allows to regenerate certificates.

This should be code reviewed in tandem with [this](https://github.com/shotgunsoftware/tk-framework-desktopserver/pull/34)

### DIFF
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -247,7 +247,8 @@ class DesktopEngineSiteImplementation(object):
         icon = QtGui.QIcon(":tk-desktop/default_systray_icon")
         app.setWindowIcon(icon)
 
-        splash.set_message("Building UI")
+        if splash:
+            splash.set_message("Building UI")
 
         # setup the global look and feel
         self._engine._initialize_dark_look_and_feel()
@@ -276,10 +277,13 @@ class DesktopEngineSiteImplementation(object):
         if kwargs.get("server") is None:
             # Initialize all of this after the style-sheet has been applied so any prompt are also
             # styled after the Shotgun Desktop's visual-style.
-            splash.set_message("Initializing browser integration.")
+            if splash:
+                splash.set_message("Initializing browser integration.")
             try:
                 desktop_server_framework = sgtk.platform.get_framework("tk-framework-desktopserver")
-                desktop_server_framework.launch_desktop_server(self._user.host, self._current_login["id"])
+                desktop_server_framework.launch_desktop_server(
+                    self._user.host, self._current_login["id"], parent=splash
+                )
             except Exception:
                 logger.exception("Unexpected error while trying to launch the browser integration:")
             else:
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -370,6 +370,18 @@ class DesktopWindow(SystrayWindow):
         self.user_menu.exec_(event.globalPos())
 
     def _show_rich_message_box(self, icon, title, message, buttons=[]):
+        """
+        Shows a QMessageBox that supports HTML formatting.
+
+        :param icon: Icon to use.
+        :type icon: ``QtGui.QMessageBox.Icon``
+        :param str title: Title of the dialog.
+        :param str message: Message to display.
+        :param list buttons: List of `QtGui.QMessageBox.StandardButton` to display.
+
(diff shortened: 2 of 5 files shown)


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -247,7 +247,8 @@ def run(self, splash, version, **kwargs):
         icon = QtGui.QIcon(":tk-desktop/default_systray_icon")
         app.setWindowIcon(icon)
 
-        splash.set_message("Building UI")
+        if splash:
```
manneohrstrom: maybe add a comment here - and in the method docstring - pointing out that this can be none and when that happens. This feels like something that is super easy to miss when writing code - is there a way we could it differently? like a `set_splash()` method or something rather than having to do the none check every single time?

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -276,10 +277,13 @@ def run(self, splash, version, **kwargs):
...
             # Initialize all of this after the style-sheet has been applied so any prompt are also
             # styled after the Shotgun Desktop's visual-style.
-            splash.set_message("Initializing browser integration.")
+            if splash:
+                splash.set_message("Initializing browser integration.")
             try:
                 desktop_server_framework = sgtk.platform.get_framework("tk-framework-desktopserver")
-                desktop_server_framework.launch_desktop_server(self._user.host, self._current_login["id"])
+                desktop_server_framework.launch_desktop_server(
+                    self._user.host, self._current_login["id"], parent=splash
```
manneohrstrom: if the splash is `None` or when the splash goes away, will the parenting here cause issues you think? Normally we deallocate in reverse order, e.g. all children are destroyed before the parent - here is suspect the desktop server will live on beyond the life cycle of the splash? In any case, may be worth a comment explaining!

python/tk_desktop/desktop_window.py
```diff
@@ -365,6 +369,85 @@ def _pop_dll_state(self):
...
+            "Shotgun browser integration",
+            "Regenerating the Shotgun Desktop's browser integration certificates should "
+            "only be done if you have issues with the browser integration.<br/>"
+            "<br/>"
+            "If you are unsure how to proceed, we recommend you visit our support page "
+            "to troubleshoot connections with <a href='%s'>Chrome</a> or "
+            "<a href='%s'>Firefox</a>.<br/>"
+            "<br/>"
+            "Would you like to continue?" % (
+                "https://support.shotgunsoftware.com/hc/en-us/articles/114094536273",
```
manneohrstrom: could these be constants?

python/tk_desktop/desktop_window.py
```diff
@@ -365,6 +369,85 @@ def _pop_dll_state(self):
...
+            self._show_rich_message_box(
+                QtGui.QMessageBox.Critical,
+                "Shotgun browser integration",
+                "It appears there are an issue while regenerating the certificates."
+                "\n"
+                "Please contact <a href='{0}'>our support team</a> "
+                "if you need assistance resolving this issue. Make sure to attach the logs found "
+                "at <a href='file://{1}'>{1}</a>.".format(
+                    "mailto:support@shotgunsoftware.com",
+                    sgtk.LogManager().base_file_handler.baseFilename
```
manneohrstrom: how about instead recommending them to zip up their entire log folder and send that? The code will be simpler but more importantly i have recently found that clients sometimes just send snippets of their log and you miss the big picture. by asking them to send all logs you get everything and there is no risk of confusion.

python/tk_desktop/desktop_engine_site_implementation.py
jfboismenu: This wouldn't be a backwards compatible change with desktop startup, we can't change this at this point.

jfboismenu: Besides, this is only meant so it is easier to bootstrap the desktop manually in PyCharm.

python/tk_desktop/desktop_engine_site_implementation.py
jfboismenu: As I mentionned in the other PR, the parenting here is so temporary pop ups while the call is being made are always displayed on top of whatever we are displaying. The framework does not keep a handle to `parent`.

PR #77, opened 2017-06-07
For #43159, Chrome 59+ fix addendum

jfboismenu:
Deactivates auto-hide behaviour in pinned mode since we'll lose focus when regenerating certificates.

It's easier to make sense of what changed if you review with this [link](https://github.com/shotgunsoftware/tk-desktop/pull/77/files?w=1) instead.

### DIFF
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -400,54 +400,62 @@ class DesktopWindow(SystrayWindow):
         """
         # Need to create the message box by hand to have rich text format, hence
         # clickable Urls.
-        choice = self._show_rich_message_box(
-            QtGui.QMessageBox.Information,
-            "Shotgun browser integration",
-            "Regenerating the Shotgun Desktop's browser integration certificates should "
-            "only be done if you have issues with the browser integration.<br/>"
-            "<br/>"
-            "If you are unsure how to proceed, we recommend you visit our support page "
-            "to troubleshoot connections with <a href='%s'>Chrome</a> or "
-            "<a href='%s'>Firefox</a>.<br/>"
-            "<br/>"
-            "Would you like to continue?" % (
-                self._CHROME_SUPPORT_URL,
-                self._FIREFOX_SUPPORT_URL
-            ),
-            [QtGui.QMessageBox.Yes, QtGui.QMessageBox.No]
-        )
 
-        if choice != QtGui.QMessageBox.Yes:
-            return
-
+        # Deactivate the auto-hide behaviour of the desktop in pinned mode when it loses focus.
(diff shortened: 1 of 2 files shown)


python/tk_desktop/desktop_window.py
```diff
@@ -400,54 +400,62 @@ def handle_regen_certs(self):
...
-            [QtGui.QMessageBox.Yes, QtGui.QMessageBox.No]
-        )
-
-        if choice != QtGui.QMessageBox.Yes:
-            return
 
+        # Deactivate the auto-hide behaviour of the desktop in pinned mode when it loses focus.
+        # since we're about to be prompted by the OS, which will make our app lose focus and
+        # hide our dialogs.
+        self.deactivate_auto_hide(True)
```
manneohrstrom:
```
with deactivated_auto_hide():
   xxx
```

just sayin.... :)

python/tk_desktop/systray.py
```diff
@@ -51,25 +51,35 @@ class ApplicationEventFilter(QtCore.QObject):
...
 
             return QtCore.QObject.eventFilter(self, obj, event)
 
+        def deactivate(self, deactivate):
+            """
+            Allows to deactivate the event filter.
+
+            :param deactivate: If ``True``, events will not be filtered anymore.
+            """
+            self._deactivate = deactivate
```
manneohrstrom: nit: is it better to call the variable `self._deacticated`?

python/tk_desktop/systray.py
```diff
@@ -51,25 +51,35 @@ class ApplicationEventFilter(QtCore.QObject):
...
+                        self._window.hide()
+                        if osutils is not None:
+                            osutils.make_app_background()
+                elif event.type() == QtCore.QEvent.ApplicationActivate:
+                    # When the app gains focus and is in pinned mode, we bring the app to the background.
+                    # Note that we are not showing the main dialog because there are multiple top levels
+                    # windows that would have cause the application to activate.
+                    if self._window.state == SystrayWindow.STATE_PINNED:
+                        if osutils is not None:
+                            osutils.make_app_foreground()
```
manneohrstrom: not sure, but would it be better to call the base class first? Not sure what it does in the case, but if it does something related to activations, it could shadow the logic in this method.

python/tk_desktop/desktop_window.py
jfboismenu: Yeah, I thought about it. :)

python/tk_desktop/systray.py
jfboismenu: We always call the base class, see right under.

PR #79, opened 2017-06-21
For #42679, can't jump into recent core

jfboismenu:
Instead of calling `prepare_engine`, we will bootstrap in the background process, just like any other DCC would do. We'll use an RPC connection to stream log messages back to the main process while bootstrapping.

The changes made in this pull request are backwards compatible with previous version of the desktop because the `bootstrap_utilities.py` is paired with the site engine and not the project engine (which might be of a different version number), which means how we communicate during bootstrap is entirely at the discretion of the site engine. @robblau , that was a great design choice you made years ago. :)

Also see [this](https://github.com/shotgunsoftware/tk-core/pull/475) to put these changes in perspective.


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -148,37 +146,31 @@ def show_panel(self, panel_id, title, bundle, widget_class,
...
+        self.site_comm.register_function(self.bootstrap_progress_callback, "bootstrap_progress")
         self.site_comm.register_function(self.engine_startup_error, "engine_startup_error")
         self.site_comm.register_function(self.set_groups, "set_groups")
         self.site_comm.register_function(self.set_collapse_rules, "set_collapse_rules")
         self.site_comm.register_function(self.trigger_register_command, "trigger_register_command")
         self.site_comm.register_function(self.project_commands_finished, "project_commands_finished")
 
     def engine_startup_error(self, error, tb=None):
-        """ Handle an error starting up the engine for the app proxy. """
-        trigger_project_config = False
```
jfboismenu: How the dialog handles errors should be handled by... the dialog.

python/tk_desktop/desktop_window.py
```diff
@@ -1188,7 +1220,8 @@ def __launch_app_proxy_for_project(self, project, requested_pipeline_configurati
 
             # If no pipeline configurations were found in Shotgun, show the
             # 'Advanced project setup...' menu item.
-            if not pipeline_configurations:
+            # If there is no pipeline configuration set for the current project, add the Advanced Project Setup menu.
+            if not any(True if pc["project"] else False for pc in pipeline_configurations):
```
jfboismenu: If a site level pipeline configuration was present, we wouldn't have been able to run `Advanced Project Setup`.

python/tk_desktop/desktop_window.py
```diff
@@ -1227,82 +1262,48 @@ def __launch_app_proxy_for_project(self, project, requested_pipeline_configurati
         # From this point on, we don't touch the UI anymore.
         self.project_overlay.start_progress()
 
-        self._sync_thread = ProjectSynchronizationThread(toolkit_manager, project)
```
jfboismenu: No more need for the sync thread, since everything is synced in the background process.

python/tk_desktop/desktop_window.py
```diff
@@ -1227,82 +1262,48 @@ def __launch_app_proxy_for_project(self, project, requested_pipeline_configurati
...
-
-    def _sync_success(self, config_path, descriptor):
         try:
-            self._current_pipeline_descriptor = descriptor
-            # Banners might need to be updated, we might have picked a configuration that has been
-            # updated.
-            self._update_banners()
-
-            engine = sgtk.platform.current_engine()
+            self._current_pipeline_descriptor = config_descriptor
```
jfboismenu: We'll update the banners after the subprocess has launched correctly now.

python/tk_desktop/desktop_window.py
```diff
@@ -1353,10 +1347,60 @@ def _sync_success(self, config_path, descriptor):
...
+                   "\n\nTo resolve this, open Shotgun in your browser\n"
+                   "and check the paths for this Pipeline Configuration."
+                   "\n\nFor more details, see the console." % message)
+        self.project_overlay.show_error_message(message)
+
+    def bootstrap_progress_callback(self, value, msg):
+        self.project_overlay.report_progress(value, msg)
+
+    def _pick_pipeline_configuration(self, pipeline_configurations, requested_pipeline_configuration_id, project):
+        """
```
jfboismenu: This isn't new code. It used to be define between two methods, but now the above method and below method have merged together so this got displaced at the bottom now.

python/utils/bootstrap_utilities.py
```diff
@@ -22,55 +22,155 @@
...
+            )
+        except Exception:
+            # If something couldn't be pickled, don't fret too much about it,
+            # we'll format it ourselves instead.
+            self._proxy.call_no_response(
+                "proxy_log", record.levelno, (record.msg % record.args), []
+            )
+
+
+class Bootstrap(object):
```
jfboismenu: The whole bootstrap sequence has been refactored into a class to avoid globals.

python/tk_desktop/desktop_window.py
```diff
@@ -298,7 +299,7 @@ def _update_banners(self):
         banner_layout = self.ui.banners.layout()
 
         # Find all the current banners.
-        current_banners = {banner_layout.itemAt(0).widget().unique_id for i in range(banner_layout.count())}
+        current_banners = {banner_layout.itemAt(i).widget().unique_id for i in range(banner_layout.count())}
```
manneohrstrom: any way this IMHO complicated line could be made more simple to read?

python/tk_desktop/desktop_window.py
```diff
@@ -1127,6 +1124,37 @@ def __get_server_version(self, connection):
...
+        """
+        Handle an error starting up the engine for the app proxy.
+
+        :param error: Exception object that was raised during bootstrap.
+        :param tb: Traceback of the exception raised during bootstrap.
+        """
+        trigger_project_config = False
+        if isinstance(error, sgtk.TankEngineInitError):
+            # match directly on the error message until something less fragile can be put in place
+            if error.message.startswith("Cannot find an engine instance tk-desktop"):
```
manneohrstrom: is this from core? Could we add a new exception type subclassing from TankEngineInitError - perhaps as part of https://github.com/shotgunsoftware/tk-core/pull/475 ?

If that is complicated, we should add a comment in core pointing out that changing that string may break desktop. But would be great if we could avoid it!

manneohrstrom: also a comment explaining a bit *why* we are doing all this could be useful i think.

python/tk_desktop/desktop_window.py
```diff
@@ -1227,82 +1258,48 @@ def __launch_app_proxy_for_project(self, project, requested_pipeline_configurati
...
             engine.startup_rpc()
 
             # pickle up the info needed to bootstrap the project python
             desktop_data = {
                 "core_python_path": core_python,
-                "config_path": config_path,
+                # Every settings that were used for discovering the pipeline configuration must be
+                # passed down to the next process so it can launch the same pipeline.
+                "manager_settings": {
+                    "caching_policy": toolkit_manager.caching_policy,
```
manneohrstrom: would it make sense that rather than picking and choosing parameters here, we added a method 
to toolkit mgr to serialize its state? That way potential new state variables could be handled gracefully? Not sure about this, curious what your thoughts are.

jfboismenu: Maybe a pickle would be good enough here actually. He know the version of core is the same on both sides here. My lack of experience with pickle with user classes makes me worried a bit. I'll go with a regular dict right now.

python/tk_desktop/desktop_window.py
```diff
@@ -1227,82 +1265,42 @@ def __launch_app_proxy_for_project(self, project, requested_pipeline_configurati
...
-            # updated.
-            self._update_banners()
-
-            engine = sgtk.platform.current_engine()
+            self._current_pipeline_descriptor = config_descriptor
+            # Find the interpreter the config wants to use.
+            try:
+                path_to_python = config_descriptor.python_interpreter
+            except TankFileDoesNotExistError:
+                if sys.platform == "darwin":
```
manneohrstrom: some debug logging here perhaps?

PR #84, opened 2017-10-17
For #44980: Update engine to use newer metric API

nmontmarquette:


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -307,12 +307,23 @@ def run(self, splash, version, **kwargs):
...
-        # websocket server.
-        sgtk.util.log_user_attribute_metric("Engine-Websockets", "no" if server else "yes")
+        try:
+            # Log usage statistics about the Shotgun Desktop executable and the desktop startup.
+            #
+            # First we update `host_info` property so subsequent metrics can benefit
+            # having the updated information. A special case is made for for Desktop
+            # as we do want both versiond but don't want to create another metric field.
+            # We are then combining both versions into single version string.
+            self._engine._host_info.update({
```
josh-t: why `update`? Should it just set the value? `self._engine._host_info["version"] = ...`

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -307,12 +307,23 @@ def run(self, splash, version, **kwargs):
...
+                "version": "%s / %s" % (self.app_version, self.startup_version)
+            })
+
+            self._engine.log_metric("Launched Software")
+
+        except Exception as e:
+            logger.exception("Unexpected error logging a metric")
```
josh-t: Should we log the error here?

python/tk_desktop/desktop_engine_site_implementation.py
nmontmarquette: We don't have the version at that time, it comes later on in the site implementation 'run' method.

python/tk_desktop/desktop_engine_site_implementation.py
nmontmarquette: The logger 'exception' method already logs a mini TraceBack, I didn't though it was useful to add anything else. Here is an example:
```
  File "/shotgun/shotgun_workspace/tk-desktop/python/tk_desktop/desktop_engine_site_implementation.py", line 321, in run
    self._engine.log_metdddric("Launched Software")
  File "/shotgun/shotgun_workspace/tk-desktop/engine.py", line 149, in __getattr__
    return getattr(self.__impl, attr)
AttributeError: 'DesktopEngineSiteImplementation' object has no attribute 'log_metdddric'
```

python/tk_desktop/desktop_engine_site_implementation.py
josh-t: Update is typically used when you have two dictionaries and you want to update one with the contents of the other. In this case, you have the host info dictionary and a value you need to set for one of its keys. In this case, `update` works but I think setting the value is better. 

```python
version = "%s / %s" % (self.app_version, self.startup_version)
self._engine._host_info["version"] = version
```

python/tk_desktop/desktop_engine_site_implementation.py
josh-t: Ah, right. Makes sense. All good then.

PR #85, opened 2017-10-21
For #43813: Adds a debug logging toggle action to the Advanced menu.

thebeeland:
The action follows these rules:

1. If TK_DEBUG is set in the environment, debug logging is always on and the menu action is disabled.
2. If the user previously used this menu action to turn debugging on or off, that is remembered in engine-level user settings and will be preserved when SG Desktop is relaunched.
3. If TK_DEBUG is not set in the environment and the user has not previously used this menu action to toggle debug logging on or off, the checked state of the action will match the global debug setting as reported by the LogManager.

The only bit here that is unfortunate is that QMenu in Qt4.x does not support showing tooltips for QActions. I've included tooltips anyway, as presumably someday this will be running in Qt 5.x and they'll just start showing properly.

### DIFF
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -51,6 +51,22 @@ class DesktopEngineSiteImplementation(object):
         shotgun_globals.unregister_bg_task_manager(self._task_manager)
         self.site_comm.shut_down()
 
+    def set_global_debug(self, state):
+        """
+        Attempts to tell a project subprocess to set the state of
+        the global debug logging setting. This will never raise
+        an exception, but a warning message will be logged if something
+        causes the RPC call to raise.
+
+        :param bool state: The debug to set.
+        """
+        try:
+            self.site_comm.call_no_response("set_global_debug", state)
+        except Exception:
+            # This really can't be a debug log call, because we might have just
+            # toggled debug logging off, in which case the message would not be
+            # logged.
+            logger.warning("RPC call to set_global_debug did not succeed.")
 
     ###########################################################################
     # panel support (displayed as tabs)
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -60,13 +60,24 @@ class DesktopEngineSiteImplementation(object):
 
         :param bool state: The debug to set.
         """
-        try:
-            self.site_comm.call_no_response("set_global_debug", state)
-        except Exception:
-            # This really can't be a debug log call, because we might have just
-            # toggled debug logging off, in which case the message would not be
-            # logged.
-            logger.warning("RPC call to set_global_debug did not succeed.")
+        if self.site_comm.is_connected:
+            try:
+                self.site_comm.call_no_response("set_global_debug", state)
+            except Exception:
+                # This really can't be a debug log call, because we might have just
+                # toggled debug logging off, in which case the message would not be
+                # logged.
+                logger.warning(
+                    "The RPC call to set_global_debug did not succeed. This is likely "
+                    "caused by an older version of the tk-desktop engine being "
+                    "used by the project. This issue can be resolved by updating to "
+                    "the latest version of tk-desktop using the 'tank updates' command."
+                )
+        else:
+            logger.debug(
+                "No connection exists to a project subprocess. No debug "
+                "toggling will occur via RPC as a result."
+            )
 
     ###########################################################################
     # panel support (displayed as tabs)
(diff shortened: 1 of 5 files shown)


python/tk_desktop/desktop_window.py
```diff
@@ -158,6 +158,16 @@ def __init__(self, parent=None):
         url_action = self.user_menu.addAction(connection.base_url.split("://")[1])
         self.user_menu.addSeparator()
         advanced_menu = self.user_menu.addMenu("Advanced")
+
+        # We will end up with a disable advanced menu item for the
+        # debug logging toggle if TK_DEBUG is set in the environment,
+        # so we want it to be a bit more obvious that the action is
+        # disabled than we get out of the box.
+        advanced_menu.setStyleSheet(
```
jfboismenu: This like we're putting the advanced_menu darker instead of the toggle. Is that on purpose? There are other things in that menu that is always availble like "Show Console..."

python/tk_desktop/desktop_window.py
```diff
@@ -170,6 +180,32 @@ def __init__(self, parent=None):
...
+        self.toggle_debug_action.toggled.connect(self._debug_toggled)
+
+        advanced_menu.addAction(self.toggle_debug_action)
+
+        # In the case of TK_DEBUG, we want to disable the control and set a tooltip 
+        # that explains why. We treat the TK_DEBUG env var as a complete override,
+        # otherwise we can check stored user settings. If the user previously set
+        # the logging on or off explicitly via this menu action then we'll remember
+        # it.
+        if sgtk.constants.DEBUG_LOGGING_ENV_VAR in os.environ:
```
jfboismenu: sgtk.constants is not part of the API. Maybe we need to introduce sgtk.LogManager().is_debug_env_var_set()? @manneohrstrom, what do you think?

manneohrstrom: If we cannot use `sgtk.LogManager().global_debug`, adding a method to core sounds useful.

python/tk_desktop/desktop_window.py
manneohrstrom: could this be in the main stylesheet file?

manneohrstrom:
wondering if we need the logic around disabling the login menu in the case `TK_DEBUG` is being set? The log manager tries to manage `TK_DEBUG` hidden under the hood. I think the idea is that as a user of the API, you should need to worry about TK_DEBUG. 

Could it be something along these lines:
- At startup, if the debug logging preference in desktop is set to True, set `global_debug` to True
- Later on, during runtime, if the checkbox is unchecked by the user, set `global_debug` to False.

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -51,6 +51,22 @@ def destroy_engine(self):
...
+        causes the RPC call to raise.
+
+        :param bool state: The debug to set.
+        """
+        try:
+            self.site_comm.call_no_response("set_global_debug", state)
+        except Exception:
+            # This really can't be a debug log call, because we might have just
+            # toggled debug logging off, in which case the message would not be
+            # logged.
```
manneohrstrom: haha! classic chicken-egg!

jfboismenu: Is it possible to know in advance if an RPC is available, or is a specific exception raised when the RPC does not exist?

In any case, the warning could suggest why the call failed. We have a pretty good idea here why it would fail. We could recommend to our users to upgrade their project's tk-desktop to `v2.2.12` to enable that feature.

python/tk_desktop/desktop_window.py
```diff
@@ -278,6 +291,48 @@ def __init__(self, parent=None):
...
+    def _debug_toggled(self, state):
+        """
+        Toggles global debug logging and stores the given state as an
+        engine level user setting.
+
+        :param bool state: The debug state to set.
+        """
+        self.user_preferred_debug_logging = state
+        sgtk.LogManager().global_debug = state
+        sgtk.platform.current_engine().set_global_debug(state)
```
jfboismenu: This will fail automatically when you are not inside a project, which will yield a warning in the console that is not warranted.

thebeeland: Resolved by adding an is_connected check lower down in the site engine implementation.

thebeeland:
Concerning toggling debug prior to browsing to a project context, that is not an issue. Because TK_DEBUG will be set (or unset) in the environment, the suprocessed spawned for the project view will inherit the setting without having to be told over the RPC connection.

HOWEVER. While I was proving that to be the case, I discovered and order of operations problem that was causing issues with exactly this case. That has been resolved in the latest commit (after JF's last comments).

PR #86, opened 2017-11-06
Splat 103 complete amplitude metrics work in toolkit

nmontmarquette:


python/tk_desktop/desktop_window.py
```diff
@@ -336,6 +342,17 @@ def _load_settings(self):
         # Update the project at the very end so the Python process is kicked off when everything
         # is initialized.
         project_id = self._settings_manager.retrieve("project_id", None, self._settings_manager.SCOPE_SITE)
+        if project_id == 0:
+            # 0 means will be displaying all of the projects
+            try:
+                from sgtk.util.metrics import EventMetric as EventMetric
+                EventMetric.log(EventMetric.GROUP_NAVIGATION,
+                                "Viewed Projects",
+                                context=sgtk.platform.current_bundle())
```
josh-t: Should probably use the `engine` here as well since this will always be a "Desktop" metric.

josh-t: Suggest `bundle` as mentioned in core PR.

python/tk_desktop/desktop_window.py
```diff
@@ -678,6 +695,16 @@ def on_project_commands_finished(self):
         """
         Invoked when all commands found for a project have been registered.
         """
+
+        try:
+            from sgtk.util.metrics import EventMetric as EventMetric
+            EventMetric.log(EventMetric.GROUP_PROJECTS,
+                            "Viewed Project Commands",
+                            context=sgtk.platform.current_bundle())
```
josh-t: Same comment about using the engine as the `bundle` arg.

python/tk_desktop/desktop_window.py
```diff
@@ -904,6 +931,14 @@ def _on_back_to_projects_clicked(self):
         # remember that we are back at the browser
         self.current_project = None
         self._save_setting("project_id", 0, site_specific=True)
+        try:
+            from sgtk.util.metrics import EventMetric as EventMetric
+            EventMetric.log(EventMetric.GROUP_NAVIGATION,
+                             "Viewed Projects",
+                             context=sgtk.platform.current_bundle())
```
josh-t: same.

PR #87, opened 2017-11-07
43269 desktop reload and restart menu item dupes

nmontmarquette:
This pull request fixes a project-related UI problem. If you select the reload and restart option from the project context menu in the top right, it will duplicate all the options in the menu.

The fix is about adding a new `proxy created hook` allowing resetting of some UI components (namely the project context menu) after engine and proxy have been reset. 

The main clear UI method is now split.

### DIFF
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -35,6 +35,7 @@ class DesktopEngineSiteImplementation(object):
 
         self.site_comm = SiteCommunication(engine)
         self.site_comm.proxy_closing.connect(self._on_proxy_closing)
+        self.site_comm.proxy_created.connect(self._on_proxy_created)
 
         self._engine = engine
         self.app_version = None
@@ -213,6 +214,11 @@ class DesktopEngineSiteImplementation(object):
         # Clear the UI, we can't launch anything anymore!
         self.desktop_window.clear_app_uis()
 
+    def _on_proxy_created(self):
+        """
+        Invoked when background process has created proxy
+        """
+        logger.debug("_on_proxy_created")
     def set_groups(self, groups, show_recents=True):
         self.desktop_window.set_groups(groups, show_recents)
 
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -218,7 +218,10 @@ class DesktopEngineSiteImplementation(object):
         """
         Invoked when background process has created proxy
         """
-        logger.debug("_on_proxy_created")
+        # Clears the project menu so the previous engine's actions
+        # are removed before adding new one
+        self.desktop_window.clear_actions_from_project_menu()
+
     def set_groups(self, groups, show_recents=True):
         self.desktop_window.set_groups(groups, show_recents)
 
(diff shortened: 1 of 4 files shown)


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -213,6 +214,15 @@ def _on_proxy_closing(self):
         # Clear the UI, we can't launch anything anymore!
         self.desktop_window.clear_app_uis()
 
+    def _on_proxy_created(self):
+        """
+        Invoked when background process has created proxy
+        """
+        logger.debug("_on_proxy_created")
+        # Clears the project menu to prevent xtra engine init from  multi-populating
```
jfboismenu: extra

jfboismenu: How about `# Clears the project menu so the previous engine's actions are removed before adding new one`

python/tk_desktop/desktop_window.py
```diff
@@ -990,13 +990,16 @@ def clear_app_uis(self):
         self.project_overlay.hide()
 
         # clear the project specific menu
+        self.clear_project_menu()
+        self.__pipeline_configuration_separator = None
+
+    def clear_project_menu(self):
```
jfboismenu: The code will have to be a bit more complicated then that. If you have multiple pipeline configurations, you will have both the pipeline configurations with their checkboxes next to them and the actions. The way you did this on Reload and Restart you will wipe all the menu. 

Here's what it looks like on my computer:

screen shot 2017-11-08 at 08 08 34

You need to removing everything up to the separator and then leave everything else in. The method can then be renamed to something more specific like `clear_engine_actions_from_project_menu`.

python/tk_desktop/project_menu.py
```diff
@@ -0,0 +1,150 @@
...
+
+log = get_logger(__name__)
+
+class ProjectMenu(object):
+    """
+    Encalsulate specific functionalities relating ot the project menu
+    """
+    def __init__(self, parent):
+        """
+
```
jfboismenu: extra empty line

python/tk_desktop/project_menu.py
```diff
@@ -0,0 +1,150 @@
...
+        :param parent: DesktopWindow-->SystrayWindow-->
```
jfboismenu: This comment looks incomplete.

python/tk_desktop/project_menu.py
```diff
@@ -0,0 +1,150 @@
...
+        """
+        self._parent = parent
+        self._project_menu = None
+        self._pipeline_configuration_separator = None
+
+    def add(self, action):
```
jfboismenu: Missing docstring

python/tk_desktop/project_menu.py
```diff
@@ -0,0 +1,150 @@
...
+        self._parent.ui.project_menu.setMenu(self._project_menu)
+
+    def clear_actions(self):
+        """
+        Clears actions from the project menu but keeps pipeline configurations if any
+        """
+        actions = self._project_menu.actions()
+        for action in actions:
+            text = action.text()
+            print text
```
jfboismenu: rogue print.

python/tk_desktop/project_menu.py
```diff
@@ -0,0 +1,150 @@
...
+            if action == self._pipeline_configuration_separator:
```
jfboismenu: A small comment explaining why we are breaking the loop here, e.g. `# the context's actions are listed above the separator, so we know we've removed all of them when we find it`

python/tk_desktop/project_menu.py
```diff
@@ -0,0 +1,150 @@
+import itertools
```
jfboismenu: missing copyright header.

PR #88, opened 2017-11-18
Adds support to tk-desktop engine for the deny_permissions option when registering commands/app launchers.

bpabel:
The base sgtk Engine already supports a ``deny_permissions`` option when registering commands, but currently, it is only supported by the tk-shotgun engine.  This adds support for the option to the tk-desktop engine as well.  This allows finer control over which SG Desktop application launchers are available based on permission roles.


bpabel:
Closing this as it's handled by using user and project restrictions now.

PR #89, opened 2017-11-26
43269 desktop reload and restart menu item dupes 2

nmontmarquette:
I was trying to understand what was creating a radio button like behavior in the code but couldn't find anything. Until I realized that the radio-button-like behavior was done by simply recreating the options all of the time and 'checking' the user choice.

There are two distinct commits in this pull request:
- Fixes a bug introduced in the last commit
- Replace UI checkboxes with actual radio buttons.

The second one is really optional and up to reviewer. The change is just a visual one. The behavior remains the same.

sg_desktop_checkboxes
sg_desktop_radio_buttons


robblau:
UI update to use radio buttons looks good.

PR #91, opened 2018-01-09
For #45989: Implements a workaround for an error reporting issue on Windows.

thebeeland:
The problem's result is that the connection from handle_error hangs on connect. The workaround is to make use of the same RPCProxy object that we already have at the time that the bootstrap is running.

The source of the problem appears to be either a client connection that doesn't actually close when it's asked to, or a server issue whereby it doesn't accept new connections when a connection is terminated. The workaround is to make use of an existing connection when we have one to use, and only try to create a new one to report errors when that's not possible. This resolves the bulk of the error reporting issues, which are most likely to occur when bootstrapping.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -46,8 +46,8 @@ def main():
         engine = utilities.start_engine(data)
         result = utilities.start_app(engine)
         os._exit(result)
-    except Exception:
-        if utilities is not None:
+    except Exception as exc:
+        if utilities is not None and not hasattr(exc, "sgtk_exception_handled"):
             # send the error back to the GUI proxy
             #
             # Use the utilities module to send the error message back to the app
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -46,8 +46,8 @@ def main():
         engine = utilities.start_engine(data)
         result = utilities.start_app(engine)
         os._exit(result)
-    except Exception as exc:
-        if utilities is not None and not hasattr(exc, "sgtk_exception_handled"):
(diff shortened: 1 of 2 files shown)


bootstrap.py
```diff
@@ -46,8 +46,8 @@ def main():
         engine = utilities.start_engine(data)
         result = utilities.start_app(engine)
         os._exit(result)
-    except Exception:
-        if utilities is not None:
+    except Exception as exc:
+        if utilities is not None and not hasattr(exc, "sgtk_exception_handled"):
```
jfboismenu: IIRC correctly, clients can take over that file, so they won't have your fix.

python/utils/bootstrap_utilities.py
```diff
@@ -239,31 +249,44 @@ def start_app(engine):
...
     if the server has already been closed (process crashes or user left the
     project) this will actually fail silently, which is alright as the main
     process doesnt't care about this one anymore.
+
+    :param proxy: An optional RPCProxy object to use when sending the
+        error to the server. If a proxy is not given, a client connection
+        will be created on the fly.
     """
+    # build a message for the GUI signaling that an error occurred
+    exc_type, exc_value, exc_traceback = sys.exc_info()
```
jfboismenu: Maybe check for your flag on the exception here intead, so the fix is self-contained to this file.

thebeeland: Cool, sounds very reasonable!

PR #92, opened 2018-01-19
For #46128, multiple endpoints in toolkit ini

jfboismenu:
While working on the new features for the multiple endpoints in toolkit.ini I found it very frustrating to debug browser integration crashes in the desktop, since these errors are not shown in the debug console, which also makes debugging harder for our clients.

I've therefore moved the initialization of Qt and the look and feel at the beggining of the `run` method. The debug console can therefore be initialized earlier and more debugging messages end up there.

Therefore, those kind of error messages will now show in the console instead of having to look at the log file.

screen shot 2018-01-18 at 10 00 48

### DIFF
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -15,7 +15,6 @@ import sys
 import string
 import collections
 
-from sgtk.platform import Engine
 from distutils.version import LooseVersion
 import sgtk
 from tank_vendor.shotgun_authentication import ShotgunAuthenticator, DefaultsManager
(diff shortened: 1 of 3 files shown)


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -342,6 +342,38 @@ def run(self, splash, version, **kwargs):
...
+        QtGui.QFontDatabase.addApplicationFont(":/tk-desktop/fonts/OpenSans-CondLight.ttf")
+        QtGui.QFontDatabase.addApplicationFont(":/tk-desktop/fonts/OpenSans-Light.ttf")
+
+        # merge in app specific look and feel
+        css_file = os.path.join(self._engine.disk_location, "resources", "desktop_dark.css")
+        with open(css_file) as f:
+            css = app.styleSheet() + "\n\n" + f.read()
+        app.setStyleSheet(css)
+
+        from .console import Console
```
josh-t: Suggest comments here saying why the console is being instantiated here instead of when it is used below. Also a comment on why the import is deferred would be nice.

PR #93, opened 2018-02-14
Fix issue with tk-desktop stomping on a specified authenticated user with the default user

mbrainerd:
This fixes an issue where the previously authenticated user (initiated during tk-framework-desktopstartup's startup.py execution) is being overridden by tk-desktop setting the current user to CoreDefaultManager's default user. This bit us when attempting to have artists switch back and forth between multiple Shotgun servers, and the connection url being incorrect since it gets pulled from the default user rather than the authenticated user.

### DIFF
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -416,16 +416,19 @@ class DesktopEngineSiteImplementation(object):
         if self.uses_legacy_authentication():
             self._migrate_credentials()
 
-        # We need to initialize current login
-        # We know for sure there is a default user, since either the migration was done
-        # or we logged in as an actual user with the new installer.
-        human_user = ShotgunAuthenticator(
-            # We don't want to get the script user, but the human user, so tell the
-            # CoreDefaultsManager manager that we are not interested in the script user. Do not use
-            # the regular shotgun_authentication.DefaultsManager to get this user because it will
-            # not know about proxy information.
-            sgtk.util.CoreDefaultsManager(mask_script_user=True)
-        ).get_default_user()
+        human_user = sgtk.get_authenticated_user()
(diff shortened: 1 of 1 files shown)


unchris:
Would this fix this stack trace?

```
  File "/Users/ccameron/Library/Caches/Shotgun/desktop/install/app_store/tk-framework-desktopstartup/v1.10.0/python/shotgun_desktop/startup.py", line 720, in main
    settings
  File "/Users/ccameron/Library/Caches/Shotgun/desktop/install/app_store/tk-framework-desktopstartup/v1.10.0/python/shotgun_desktop/startup.py", line 334, in __launch_app
    return __post_bootstrap_engine(splash, app_bootstrap, engine, settings)
  File "/Users/ccameron/Library/Caches/Shotgun/desktop/install/app_store/tk-framework-desktopstartup/v1.10.0/python/shotgun_desktop/startup.py", line 488, in __post_bootstrap_engine
    startup_descriptor=startup_desc
  File "/Applications/Shotgun.app/Contents/Resources/Python/bundle_cache/app_store/tk-desktop/v2.4.12/python/tk_desktop/desktop_engine_site_implementation.py", line 467, in run
    self.desktop_window = desktop_window.DesktopWindow(console)
  File "/Applications/Shotgun.app/Contents/Resources/Python/bundle_cache/app_store/tk-desktop/v2.4.12/python/tk_desktop/desktop_window.py", line 170, in __init__
    "HumanUser", [["id", "is", user["id"]]],
TypeError: 'NoneType' object has no attribute '__getitem__'
```

It appears that `user` on line 170 is `None` after calling `user = engine.get_current_login()`. Based on the comments about it being about switching between accounts, my guess is this bit me because I had originally tried to sign into an incorrect URL, and then tried again with the correct one. Perhaps it then pulled `None` from the previous failed attempt to connect?

PR #94, opened 2018-03-12
For #46567, background download of zip

jfboismenu:
Ensures that a configuration is local before trying to use it. It spawns off a thread first so that there is no UI freeze when the project is selected.

This also removes unused files from the tk-desktop package. It needs to actually move some files around in the repo so the `resources` folder can be easily discarded.


python/utils/bootstrap_utilities.py
```diff
@@ -223,9 +223,12 @@ def start_app(engine):
         app.setApplicationName("%s Python" % engine.context.project["name"])
 
         # set default icon
-        python_icon = os.path.realpath(os.path.join(
-            os.path.dirname(__file__),
-            "..", "..", "resources", "python_icon.png"))
+        python_icon = os.path.realpath(
```
josh-t: I'd suggest putting icons in the repo's root in an `icons` folder. Having them in `python/utils` doesn't feel right.

python/tk_desktop/desktop_window.py
```diff
@@ -1260,8 +1292,30 @@ def __launch_app_proxy_for_project(self, project, requested_pipeline_configurati
         # From this point on, we don't touch the UI anymore.
         self.project_overlay.start_progress()
 
+        self.project_overlay.report_progress(0.05, "Downloading configuration.")
+        self._current_download_thread = ConfigDownloadThread(
```
josh-t: What UI is available while this is happening in the bg? Anything that could get the user in a weird state? For example, is the context menu available? Just curious.

jfboismenu: Yes, the context menu is indeed available but only shows the different pipeline configurations accessible at this point. Just like before you can switch pipeline configuration if you want.

I've added a few comments to the code however to clarify the story around threading.

python/tk_desktop/desktop_engine_project_implementation.py
```diff
@@ -251,9 +251,9 @@ def _initialize_application(self):
         # Make the name pretty for the tray and the task manager.
         app.setApplicationName("%s Python" % self._engine.context.project["name"])
         # set default icon
-        python_icon = os.path.realpath(os.path.join(
-            os.path.dirname(__file__),
-            "..", "..", "resources", "python_icon.png"))
+        python_icon = os.path.join(
+            self._engine.disk_location, "icon_bg_256.png"
```
josh-t: minor: suggest leaving the icon name as `python_python.png` or maybe `icon_python_proc.png`.

PR #97, opened 2018-06-19
For SG-6816, jump to project fix

jfboismenu:
When clearning the menu, do not remove the Jump To xxx action. This one gets turned on or off on demand.

Note that this only fixes the issue for classic configs. Bootstraped configs are running into an issue. 

The issue is simple. A path cache parent folder for a plugin based config is `p<project-id>c<pc-id>.<plugin-name>`. This means that every plugin has it's own copy of the path cache, therefore each plugin sync their updates in their own local path cache. Since desktop does not sync the path cache on startup, it is always empty, so the menu item will never appear.

### DIFF
--- a/python/tk_desktop/project_menu.py
+++ b/python/tk_desktop/project_menu.py
@@ -68,9 +68,11 @@ class ProjectMenu(object):
 
         # Loop through actions and delete the ones listed before the separator
         for action in actions:
-            if action in [
-                self._pipeline_configuration_separator, self._parent.ui.actionProject_Filesystem_Folder
-            ]:
+            # Do not delete the Jump To Filesystem menu. This one is hidden on demand instead.
+            if action == self._parent.ui.actionProject_Filesystem_Folder:
+                continue
+
+            if action == self._pipeline_configuration_separator:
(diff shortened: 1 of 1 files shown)


python/tk_desktop/project_menu.py
```diff
@@ -68,6 +68,10 @@ def clear_actions(self):
 
         # Loop through actions and delete the ones listed before the separator
         for action in actions:
+            # Do not delete the Jump To Filesystem menu. This one is hidden on demand instead.
+            if action == self._parent.ui.actionProject_Filesystem_Folder:
```
manneohrstrom: i realise this is a small fix, but wondering if there is something more toolkit-onic we can do to solve it? So that the app decides whether the action should be shown, not the engine. The engine shouldn't know about the apps - if i am understanding this correctly, we are hard coding an app into the engine code, right? Could we use the event system that nuke studio uses to signal? Could we be smart about registering commands so that the app doesn't register the menu entry (command) if the path cache isn't there?

manneohrstrom: JF explained that this is all in the engine - which confused me! That means it's all good to go!

PR #99, opened 2019-04-09
Adjusted timing of overlay hiding before apps are shown, so they aren't erroneously highlighted.

raphaelmatto-wdi:
Hey tkFolk,

Here's a quick/dirty fix to the issue of apps that are erroneously highlighted during project load. See this WDI support ticket: https://support.shotgunsoftware.com/hc/en-us/requests/99803.

The only functional downside is: if users have the "Show Details" panel open, it very briefly is displayed on top of the apps as they are loading.

I'm sure there is a more elegant solution, but this works & is a small code change; I hope it prompts a "real" fix from someone who is more familiar with the app :)

Cheers,
Raphe

### DIFF
--- a/python/tk_desktop/desktop_engine_project_implementation.py
+++ b/python/tk_desktop/desktop_engine_project_implementation.py
@@ -79,6 +79,7 @@ class DesktopEngineProjectImplementation(object):
             self._connect_to_server()
             self._register_groups()
             self._register_commands()
+            self._project_comm.call("hide_overlay")
         except:
             # Same deal as during init_engine.
             self.destroy_engine()
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -189,6 +189,7 @@ class DesktopEngineSiteImplementation(object):
         self.site_comm.register_function(self.bootstrap_progress_callback, "bootstrap_progress")
         self.site_comm.register_function(self.engine_startup_error, "engine_startup_error")
         self.site_comm.register_function(self.set_groups, "set_groups")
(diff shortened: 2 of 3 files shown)


raphaelmatto-wdi:
Image for reference:

desktopFix

raphaelmatto-wdi:
Image, for reference:

desktopFix2

j0yu:
Hey @raphaelmatto-wdi , does this PR still work for you if it is applied to the latest ShotgunSoftware v2.4.13 release?

I'm running into issues where the `hide_overlay` doesn't get triggered by RPC which I though was bizzare see [v2.4.13+wwfx.1.0.0](https://github.com/wwfxuk/tk-desktop/releases/tag/v2.4.13%2Bwwfx.1.0.0). I just hot fixed it on [v2.4.13+wwfx.1.0.1](https://github.com/wwfxuk/tk-desktop/releases/tag/v2.4.13%2Bwwfx.1.0.1) by explicitly calling `self.desktop_window.project_overlay.hide()` at the end of `project_commands_finished` to achieve the same effect

raphaelmatto:
Hi @j0yu , yes we ran into issues as well & have modified this a bunch since this PR. If SG folks are amendable to the possibility of adding it, we might be able to post the new code. It takes care of a bunch of random issues we ran into in production w/folks on different versions of Desktop. I'll check internally if we can post it anyway.

jfboismenu:
Hi everyone, there's big problems with the project commands panel code in PySide2, which we have to support in order to support Python 3 with Desktop. This lead to a complete rewrite of the project commands panel. In the upcoming version, there's will not be any more highlight issues. We'll close this PR.

PR #106, opened 2020-02-21
SG-14617: python 3 port part 1

jfboismenu:
With these changes, it is now possible to launch a pipeline configuration that uses Python 3 with the Shotgun Desktop. While the Shotgun Desktop itself only runs on Python 2, any project can be configured to use Python 3.

To do this, we've had to completely reimplement the RPC layer between the two Desktop processes because the `multiprocessing.connection` module forced version 3 of the pickle protocol and the Shotgun Desktop, which uses Python 2.7, only understands protocols 2 and below. We've therefore implemented an HTTP based protocol, which unfortunately is a bit slower but at least works across Python versions.

Note: Disregard the commit history on this PR, it is misleading. I've forked some of the changes to another branch to keep this one focused on a single change, i.e. being able to launch a project that uses Python 3.


robblau:
Nice work. That codecov report is awesome.

python/tk_desktop/rpc.py
```diff
@@ -48,19 +149,19 @@ class RPCServerThread(threading.Thread):
...
-    def __init__(self, engine):
+    def __init__(self, engine, authkey=None):
         threading.Thread.__init__(self)
 
         # registry for methods to call for names that come via the connection
         self._functions = {
             "list_functions": self.list_functions,
         }
 
-        self._stop = False  # used to shut down the thread cleanly
```
jfboismenu: _stop is an attribute from the base class on Python 3!

python/tk_desktop/rpc.py
```diff
@@ -105,8 +205,9 @@ def run(self):
         Run the thread, accepting connections and then listening on them until
         they are closed.  Each message is a call into the function table.
         """
-        logger.debug("server listening on '%s'", self.pipe)
-        while True:
+        logger.debug("multi server listening on '%s'", self.pipe)
+        while self._is_closed is False:
```
jfboismenu: The tests exposed a situation where the server wouldn't shut down if the closed flag was raised while there wasn't a client available, so I've fixed this.

python/tk_desktop/rpc.py
```diff
@@ -131,22 +233,21 @@ def run(self):
...
-            # connection waiting to be read, accept it
-            connection = self.server.accept()
-            logger.debug("server accepted connection")
             try:
-                while True:
+                # connection waiting to be read, accept it
+                logger.debug("multi server about to accept connection")
+                connection = self.server.accept()
+                logger.debug("multi server accepted connection")
+                while self._is_closed is False:
```
jfboismenu: When coming at the top of the loop, there's no need to poll if the port is closed. This allows us to use smaller timeouts in the tests.

python/tk_desktop/rpc.py
```diff
@@ -219,41 +334,48 @@ def __init__(self, pipe, authkey):
...
         )
-        logger.debug("client connected to %s", pipe)
+        logger.debug("multi client connected to %s", pipe)
 
     def call_no_response(self, name, *args, **kwargs):
-        msg = "client calling '%s(%s, %s)'" % (name, args, kwargs)
+        msg = "multi client calling '%s(%s, %s)'" % (name, args, kwargs)
         if self._closed:
-            raise EOFError("closed " + msg)
+            raise RuntimeError("closed " + msg)
```
jfboismenu: No-one was listening for these specific errors. It makes more sense to use RuntimeError's here as we're not overloading an existing exception type and it helps discern unexpected errors from expected ones simply by looking at the exception type.

python/tk_desktop/rpc.py
```diff
@@ -219,41 +334,48 @@ def __init__(self, pipe, authkey):
...
 
         # wait until there is a result, pause to check if we have been closed
         while True:
-            if self._connection.poll(self.LISTEN_TIMEOUT):
-                # have a response waiting, grab it
-                break
-            else:
-                # no response waiting, see if we need to stop the client
+            try:
+                if self._connection.poll(self.LISTEN_TIMEOUT):
```
jfboismenu: On Windows, an IOError is raised when the connection is closed during polling instead of returning `False`.

python/tk_desktop/project_communication.py
```diff
@@ -29,13 +34,21 @@ def __init__(self, engine):
...
-        Sets up a server to communicate with the background process and connects to the site engine.
+        Sets up a server to communicate with the foreground process and connects
+        back to the site engine.
         """
         # create the connection to the site engine.
         self._create_proxy(pipe, auth)
 
-        # register our side of the pipe as the current app proxy
-        self._create_server()
+        # register our side of the pipe as the current app proxy.
```
pscadding:
```suggestion
        # Register our side of the pipe as the current app proxy.
```

python/tk_desktop/project_communication.py
```diff
@@ -29,13 +34,21 @@ def __init__(self, engine):
...
+        # Based on how the server expects us to connect to them, we'll
+        # use the right server to listen to their requests.
+        # This boils down to if the server suggested we use multiprocessing
+        # to connect back to them, we'll use that to receive their message
+        # since it's likely they are using that on their end as well.
+        # If the suggested we use http to connect back to them, then we'll listen
```
pscadding: I think this sentence is missing something, it doesn't quite read correctly.

pscadding: Also if I understand this correctly, both the project side and the site side are registering servers, so that both can receive instructions.

Based on that understanding, I'm not sure the comment is clear:
> Based on how the server expects us to connect to them, we'll use the right server to listen to their requests.

Which server? I think this should maybe be:
"Based on how the site server expects us to connect to them, we'll use the right server to listen to their requests."

python/tk_desktop/site_communication.py
```diff
@@ -42,11 +43,15 @@ def _create_proxy(self, pipe, authkey):
...
+    @property
+    def server_pipes(self):
+        return self._msg_server.pipes
+
     def start_server(self):
         """
         Sets up a server to communicate with the background process.
         """
-        self._create_server()
+        self._create_server(DualRPCServer)
```
pscadding: Ah is this why the `_create_server` is setup to be passed a server factory object, since it needs to be able to be passed this class?

python/utils/bootstrap_utilities.py
```diff
@@ -128,7 +139,7 @@ def start_engine(self):
             manager.restore_settings(self._manager_settings)
             manager.pre_engine_start_callback = self._pre_engine_start_callback
             manager.progress_callback = self._progress_callback
-
+            raise RuntimeError("Oh SNAP!")
```
pscadding: is this left over from testing? This doesn't seem right, as the following lines will never be reached.

tests/test_rpc.py
```diff
@@ -0,0 +1,412 @@
...
+
+
+@pytest.fixture
+def proxy(server):
+    """
+    Creates a proxy that can communicate with the server fixture.
+
+    :returns: A proxy instance.
+    """
+    client = server.client_factory(server)
```
pscadding: I find this very confusing.
`server` is a method/fixture in the global scope, and that method creates and yields a variable called `server` which contains a `client_factory` callable that wants to be passed `server`. 😕
I think maybe I get it?

tests/test_rpc.py
```diff
@@ -0,0 +1,412 @@
...
+    with contextlib.closing(MultiprocessingRPCServerThread(fake_engine)) as server:
+        server.start()
+        with pytest.raises(Exception) as exc:
+            MultiprocessingRPCProxy(server.pipe, b"12345")
+        assert str(exc.value) == "digest sent was rejected"
+
+
+def test_calling_when_closed(proxy):
+    """
+    Ensure calling a method on the clietn when it is closed fails.
```
pscadding:
```suggestion
    Ensure calling a method on the client when it is closed fails.
```

python/utils/bootstrap_utilities.py
pscadding: Ha, I see you removed it just as I posted my review 😆

python/tk_desktop/communication_base.py
```diff
@@ -101,24 +101,20 @@ def _create_proxy(self, pipe, authkey):
...
-        self._proxy = RPCProxy(pipe, authkey)
-        logger.debug("Connected to the proxy server.")
+        self._proxy = get_rpc_proxy_factory(pipe)(pipe, authkey)
+        logger.info(
+            "%s connected to proxy server using %s(%s)."
+            % (type(self).__name__, type(self._proxy).__name__, pipe)
+        )
 
-    def _create_server(self):
+    def _create_server(self, rpc_server_factory):
```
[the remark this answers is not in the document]
jfboismenu: That's because on the SiteCommunication side, we always use the DualRPCServer class, since we don't know how the background desktop will connect. On the ProjectCommunication side however, we'll use the http proxy or the multiprocessing proxy, depending on which is available.

python/tk_desktop/desktop_engine_project_implementation.py
```diff
@@ -91,10 +91,11 @@ def _connect_to_server(self):
         # pull the data on how to connect to the GUI proxy from the tk instance
         bootstrap_data = self._engine.sgtk._desktop_data
         proxy_pipe = bootstrap_data["proxy_pipe"]
+        http_pipe = bootstrap_data.get("http_pipe")
         proxy_auth = bootstrap_data["proxy_auth"]
 
         self._project_comm.connect_to_server(
-            proxy_pipe, proxy_auth, self._signal_disconnect
+            http_pipe or proxy_pipe, proxy_auth, self._signal_disconnect
```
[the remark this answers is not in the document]
jfboismenu: Good point!

python/tk_desktop/site_communication.py
jfboismenu: Yup!

python/tk_desktop/project_communication.py
jfboismenu: I rewrote the blurb, let me know what you think!

tests/test_rpc.py
jfboismenu: I added a comment 😄

python/tk_desktop/rpc.py
```diff
@@ -8,29 +8,128 @@
...
+HttpRPCProxy.
+
+The Http based implementation is a bit more complicated and involves a few
+more classes internally. The relationship between the classes are as follow:
+
+- HttpRPCServerThread owns the Listener.
+- The Listener is used to register methods and owns an HTTPServer that the Connection
+  will invoke.
+- The HttpRPCProxy owns a Connection object to send data to the HTTPServer.
+"""
```
pscadding: Great description!

python/tk_desktop/rpc.py
```diff
@@ -8,29 +8,128 @@
...
 
-# Only log debug messages if they are specifically requested.
-if "TK_DESKTOP_RPC_DEBUG" in os.environ:
-    logger.setLevel(logging.DEBUG)
-else:
-    logger.setLevel(logging.INFO)
+class Logger(object):
+    """
+    Wrapper around a logger object. It augments the logged information when
+    TK_DESKTOP_RPC_DEBUG is turned on.
```
pscadding: Why does this only handle `debug`, `error`, `exception` and `warning` and not `info` as well?

The comment here clearly states this is only used when debugging is enabled so I assume this is intentional. Scanning the code, it seem you are only logging debug messages within the rpc module. So I assume this logging object is only for use within the rpc module and does not affect normal logging coming from say an app running within a project process.

python/tk_desktop/rpc.py
```diff
@@ -117,6 +216,7 @@ def run(self):
                     )
                     ready = True
```
pscadding: I know it's not one of your changes and I suspect the likely hood of this being an issue is low, but I think there is a chance that `ready` might not get defined, if it fails in the `try` with an error other than `WindowsError`.

python/tk_desktop/rpc.py
```diff
@@ -191,16 +293,31 @@ def run(self):
                 # just keep serving new connections
                 pass
             finally:
-                logger.debug("server closing")
-                connection.close()
+                # It's possible we failed accepting, so the variable may not be defined.
+                if "connection" in locals():
```
pscadding: Why is it being done like this, rather than say defining `connection = None` just before the start of the `try` at line `236`?

python/tk_desktop/rpc.py
```diff
@@ -219,41 +336,49 @@ def __init__(self, pipe, authkey):
...
-            else:
-                # no response waiting, see if we need to stop the client
+            try:
+                if self._connection.poll(self.LISTEN_TIMEOUT):
+                    # have a response waiting, grab it
+                    break
+                else:
+                    # If the connection was closed while we polled, raise an error.
+                    if self._closed:
+                        raise RuntimeError("client closed while waiting for a response")
```
pscadding: I noticed that you altered a previous message to become `multi client ...` should this also be updated as well as the raise a few lines below?

python/tk_desktop/project_communication.py
pscadding: Yeah I think that makes it clearer. But since this is in the `ProjectComunication` class, I still think it doesn't hurt to say site server when talking about the server's expectations, since this isn't a generic class that could be used on either side of the conversation.

> If the suggested we use http to connect back to them, then we'll listen on that as well.

as well or instead? I think instead?
Sorry being a bit picky, I think it looks better though!

python/tk_desktop/rpc.py
```diff
@@ -268,3 +393,540 @@ def close(self):
...
+    def register_function(self, func, name=None):
+        """
+        Register a function with the server.
+
+        :param callable func: Method to call.
+        :param name: Optional. If set, the function will be named as the parameter states.
+            Otherwise, func.__name__ will be used.
+        """
+        self._multiprocessing_thread.register_function(func, name)
+        self._http_thread.register_function(func, name)
```
pscadding: I'm sure this is fine, but since this code is handling both multiprocessing and http servers, do we increase the risk that if one of them errors for some reason, everything will stop even if we weren't using that type of server?

python/tk_desktop/rpc.py
```diff
@@ -268,3 +393,540 @@ def close(self):
...
+            response = e
+            # if any of the above fails send the exception back to the client
+            logger.exception("got exception during '%s" % func_name)
+        finally:
+            self._send_response(response)
+
+    def log_request(self, code=None, size=None):
+        """
+        Prevent the server from outputting to stdout every single time
+        a query is processed by the server.
```
pscadding: As I understand it, this method is overridden and intentionally left blank so that it doesn't output everytime a query is processed. 
I did a double take when looking at this, might be good to tweak the description to explicitly state that it was left empty to prevent thee server from ...

python/tk_desktop/rpc.py
```diff
@@ -268,3 +393,540 @@ def close(self):
...
+        except Exception:
+            raise RuntimeError("Response could not be unserialized: %s" % body)
+
+    def _send_response(self, body):
+        """
+        Send a response back to the client.
+
+        :param body: Data to send back to the client.
+        """
+        raw_data = pickle.dumps(body)
```
pscadding: Are we sufficiently protecting the code from loading any rogue/malicious content?
I assume that the authentication is enough to ensure that we are talking to who we think we are?

python/tk_desktop/rpc.py
```diff
@@ -268,3 +393,540 @@ def close(self):
...
+        :returns: The data that was sent.
+        """
+        if six.PY3:
+            content_len = int(self.headers.get("content-length", 0))
+        else:
+            content_len = int(self.headers.getheader("content-length", 0))
+
+        body = self.rfile.read(content_len)
+        try:
+            return pickle.loads(body)
```
pscadding: Are we sufficiently protecting the code from loading any rogue/malicious content?
I assume that the authentication is enough to ensure that we are talking to who we think we are?

python/tk_desktop/rpc.py
```diff
@@ -268,3 +393,540 @@ def close(self):
...
+        Call a function on the server synchronously or not.
+
+        :param bool is_blocking: If ``True``, the call will block until the server
+            completes the task and the value/exception will be returned/raised by this
+            method. If ``False``, the call will be queued for delivery and the method
+            returns None.
+        """
+        payload = [is_blocking, func_name, args, kwargs]
+        payload = pickle.dumps([self._authkey, payload])
+        # if is_blocking:
```
pscadding: is this comment supposed to be here?

python/tk_desktop/rpc.py
```diff
@@ -268,3 +393,540 @@ def close(self):
...
+        request = urlopen(self._address, data=six.ensure_binary(payload))
```
pscadding: Is this line adding it to the queue?

I'm clear on the non blocking behaviour here, it seems we create a request and then close it straight away?

python/tk_desktop/rpc.py
jfboismenu: I should have provided a bit more context. As far as the code review is concerned, TK_DESKTOP_RPC_DEBUG already existed before. I simply made it better.

I'll update the code.

jfboismenu: I forgot about `info`!

python/tk_desktop/rpc.py
jfboismenu: Good point. Fixed.

python/tk_desktop/rpc.py
jfboismenu: I think the technical term for doing things this way is called a "brainfart".

python/tk_desktop/rpc.py
jfboismenu: This message is protocol agnostic and the same for both protocols, so I'd leave it as is.

python/tk_desktop/rpc.py
jfboismenu: If this breaks, then I've written some really terrible tests. 😜 

Besides, if registering a function fails, then the Desktop will crash and give a pop-up.

python/tk_desktop/rpc.py
jfboismenu: For someone to be able to connect to the port, they need to know the auth key and the exact port number to connect to. Yes, it's a risk, but one we've always lived with before.

python/tk_desktop/rpc.py
jfboismenu: nope!

python/tk_desktop/rpc.py
jfboismenu: I've cleared up the confusion.

jfboismenu:
I think I've addressed all your comments. Let me know when you think it's good for QA.

PR #108, opened 2020-04-30
SG-15718: fixes panel on Windows

jfboismenu:
The previous code caused flickering on Windows. Now we were computing the sizes ourselves, which solves the issue.

### DIFF
--- a/python/tk_desktop/command_panel/recent_list.py
+++ b/python/tk_desktop/command_panel/recent_list.py
@@ -22,7 +22,6 @@ class RecentList(BaseIconList):
 
     def __init__(self, parent):
         super(RecentList, self).__init__(parent, QtGui.QHBoxLayout())
-        self.setLayout(self._layout)
         self._layout.addStretch(1)
 
     def add_command(self, command_name, button_name, icon, tooltip, timestamp):
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -1025,7 +1025,6 @@ class DesktopWindow(SystrayWindow):
         self._command_panel.add_command(
             name, button_name, menu_name, icon, command_tooltip, groups, is_menu_default
         )
-        self._force_panel_resize()
         self._project_command_count += 1
 
     def _handle_project_data_changed(self):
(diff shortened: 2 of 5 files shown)


python/tk_desktop/command_panel/recent_list.py
```diff
@@ -22,7 +22,6 @@ class RecentList(BaseIconList):
 
     def __init__(self, parent):
         super(RecentList, self).__init__(parent, QtGui.QHBoxLayout())
-        self.setLayout(self._layout)
```
jfboismenu, line 25: This line was present in the base class.

python/tk_desktop/desktop_window.py
```diff
@@ -1081,28 +1080,6 @@ def set_groups(self, groups, show_recents=True):
...
-
-    def _force_panel_resize(self):
-        """
-        Force the command panel area to recompute the size of
-        its children.
-        """
-        # On Windows, we need to force a geometry update
-        # and widget resize so we do not get a horizontal
-        # scroll in the command panel.
-        self.ui.command_panel_area.updateGeometry()
```
jfboismenu, line 1103: This code was causing the flickering.

tests/run_command_panel.py
```diff
@@ -92,7 +111,7 @@ def save(self, key, recents):
     def load(self, key):
         return {
             command: {"timestamp": datetime.datetime.now()}
-            for command, _, _, _, _, _, _ in list(commands)[0:3]
+            for command, _, _, _, _, _, _ in list(commands)[0:6]
```
pscadding, line 114: This line is very hard to read, I know you only upped a number, but I don't know why. It's picking the first six values out of a commands variable that I think was a dictionary  but is being cast to a list.

pscadding: and that contains 7 values to be unpack of which we only want the first one.

pscadding: feel free to ignore, but just reading this, it hard to understand in what way it is parsing the data structure.

tests/run_command_panel.py
```diff
@@ -132,19 +152,18 @@ def add_button():
...
-                    "from PySide2 import QtWidgets; QtWidgets.QApplication([]).exec_()",
-                ]
-            )
+            # subprocess.Popen(
+            #     [
+            #         sys.executable,
+            #         "-c",
+            #         "from PySide2 import QtWidgets; QtWidgets.QApplication([]).exec_()",
+            #     ]
+            # )
```
pscadding, line 161: Should this be commented out still?

PR #109, opened 2020-05-05
SG-17383: fixes interprocess communication

jfboismenu:
The implementation of the HTTP-based implementation was completely reverted. Before the Python 3 port, we used `multiprocessing.connection` to communicate between processes.

During the port, it came to light that Python 3 would use the version 3 of the protocol to serialize data and that a Python 3 process wouldn't be able to unpickle the data. It was incorrectly assumed that we could override this behaviour and a complicated HTTP-based implementation was created and lots of tests were written. All seemed well.

Unfortunately, hours after deploying the new solution support tickets coming in because the HTTP-based implementation was understandably wrecking havoc with HTTP and TCP proxies. This new design ended being a terrible technical decision (by yours truly!).

After careful reviewing on the Python implementation of `multiprocessing.connection`, which clearly had not been done closely enough, the `Connection` object has `recv_bytes` and `send_bytes`, which send data to the other side and `recv`/`send` methods, which the original implementation used, that pickle/unpickle objects in conjonction with `recv_bytes` and `send_bytes`.

As such, the implementation of the `Connection` has been wrapped in `SafePickleConnection` which catches call to `send`/`recv` and makes sure data can be marshalled between Python 2 and 3.

Therefore, a LOT of code from rpc.py has been reverted.


python/tk_desktop/communication_base.py
```diff
@@ -12,7 +12,7 @@
 Implements communication channels between the desktop app and the background process.
 """
 
-from .rpc import get_rpc_proxy_factory
+from .rpc import RPCServerThread, RPCProxy
```
jfboismenu, line 15: This reverts most changes made originally when implementing Python 3 support.

python/tk_desktop/desktop_engine_project_implementation.py
```diff
@@ -98,7 +98,7 @@ def _connect_to_server(self):
 
         # We always prefer the HTTP pipe, as it works under every Python version.
         self._project_comm.connect_to_server(
-            http_pipe or proxy_pipe, proxy_auth, self._signal_disconnect
+            proxy_pipe, proxy_auth, self._signal_disconnect
```
jfboismenu, line 101: This reverts most changes made originally when implementing Python 3 support.

python/tk_desktop/desktop_window.py
```diff
@@ -1489,9 +1489,8 @@ def _start_bg_process(self, config_descriptor, toolkit_manager):
                 "project": self.current_project,
                 # Authentication credentials to connect back to this process.
                 "proxy_data": {
-                    "proxy_pipe": engine.site_comm.server_pipes[0],
+                    "proxy_pipe": engine.site_comm.server_pipe,
```
jfboismenu, line 1492: This reverts most changes made originally when implementing Python 3 support.

python/tk_desktop/project_communication.py
```diff
@@ -40,17 +39,8 @@ def connect_to_server(self, pipe, auth, disconnect_callback):
...
-        # or http based, we'll create the right server to listen to request
-        # from the main desktop process.
-        #
-        # The logic is this: If the main desktop process supports http,
-        # we'll always use that to connect back because this is the safest
-        # way to exchange data. If the main desktop process didn't support http
-        # then the pipe will be a file path and therefore get_rpc_server_factory
-        # will return a multiprocessing-based server.
-        self._create_server(get_rpc_server_factory(pipe))
+        # register our side of the pipe as the current app proxy
```
jfboismenu, line 42: This reverts most changes made originally when implementing Python 3 support.

python/tk_desktop/rpc.py
```diff
@@ -44,12 +44,40 @@
 # type that is stored in the pickle and assumes the output of a dump
 # can always be turned into a utf-8 encoded string, which is false.
 # Therefore, we'll simply take the raw pickle and send it over the wire.
-from sgtk.util import pickle
-from tank_vendor.six.moves.BaseHTTPServer import HTTPServer
-from tank_vendor.six.moves.SimpleHTTPServer import SimpleHTTPRequestHandler
-from tank_vendor.six.moves.urllib.request import urlopen
 from tank_vendor import six
 from tank_vendor.six.moves import cPickle as py_pickle
+from tank.util import pickle as tk_pickle, is_windows
```
jfboismenu, line 49: This pretty much reverts the file to the state it was before the Python 3 port. What I've done in this branch is basically restore the original file and report it to Python 3.

python/tk_desktop/rpc.py
```diff
@@ -397,8 +455,14 @@ def call(self, name, *args, **kwargs):
                 raise
 
         # read the result
-        result = pickle.loads(self._connection.recv())
-        logger.debug("multi client got result '%s'" % result)
+        try:
+            result = pickle.loads(self._connection.recv())
+        except OSError:
+            # On Linux, an exception will be raised here instead.
```
jfboismenu, line 461: Using multiprocessing under Python 3 introduced some behaviour change, which you can see here.

python/tk_desktop/rpc.py
```diff
@@ -413,550 +477,3 @@ def close(self):
         logger.debug("closing connection")
         self._connection.close()
         self._closed = True
-
-
-class HttpRPCServerThread(threading.Thread):
```
jfboismenu, line 418: All that complex code i wrote, GONE!

python/tk_desktop/site_communication.py
```diff
@@ -12,7 +12,6 @@
 Implements communication channels between the desktop app and the background process.
 """
 
-from .rpc import DualRPCServer
```
jfboismenu, line 15: This reverts most changes made originally when implementing Python 3 support.

python/utils/bootstrap_utilities.py
```diff
@@ -34,6 +34,7 @@ class ProxyLoggingHandler(logging.Handler):
     def __init__(self, proxy):
         """
         :param proxy: Connection to the main process.
+        :type proxy: rpc.RPCProxy
```
jfboismenu, line 37: This reverts most changes made originally when implementing Python 3 support.

tests/test_rpc.py
```diff
@@ -126,50 +120,21 @@ def fake_engine():
...
 
 
-# We need to test the Http and Multiprocessing proxy connecting to each server
-# type.
-# Note that on Python 3, the multiprocessing server is never used by Desktop,
-# so we can ignore those servers.
-@pytest.fixture(
-    params=(
-        [
-            (DualRPCServer, MultiprocessingRPCProxy),
```
jfboismenu, line 136: We only have one implementation now, so the tests are simplified a lot.

python/tk_desktop/rpc.py
jfboismenu: You're quite right actually. I did revert to the original file. When I typed this I had forgotten that I had reversed course and restored to the ported code and stripped things out instead. This is because while writing the tests I had actually found some bugs in the RPC code, so I kept those fixes in this final version and simply tweaked about the recv/send operations were done.

python/tk_desktop/rpc.py
```diff
@@ -185,7 +243,7 @@ def __init__(self, engine, authkey=None):
         else:
```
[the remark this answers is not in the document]
jfboismenu: Doesn't matter actually, since we're doing ensure_binary later on. I just removed that.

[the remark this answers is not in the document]
pscadding: Yeah I saw that, but is it important that it gets encoded before being added to the `self.authkey` attribute. Are there other places that would use the `self.authkey` attribute and also have to remember to encode it?

PR #111, opened 2020-05-22
SG-17545 pre v2.5.3 PySide fix

pscadding:
Adds a fix to allow projects that use a version of `tk-desktop` pre `v2.5.3` to load Qt based apps.

It does this by patching the PySide2 libraries to act as PySide, and then inserting them into `sys.modules`.
It only does this if PySide or PyQt4 don't already exist in the environment.

### DIFF
--- a/python/utils/bootstrap_utilities.py
+++ b/python/utils/bootstrap_utilities.py
@@ -223,10 +223,10 @@ class Bootstrap(object):
         """
         # First try to see if PySide or PyQt4 exist in the current environment before attempting to patch sys modules.
         try:
-            import PySide
+            import PySide  # noqa
         except ImportError:
             try:
-                import PyQt4
+                import PyQt4  # noqa
             except ImportError:
                 # PySide and PyQt4 don't exist so we need to patch the sys.modules to
                 # add PySide2 as if it was PySide. This is so if we are using a
--- a/python/utils/bootstrap_utilities.py
+++ b/python/utils/bootstrap_utilities.py
@@ -148,6 +148,11 @@ class Bootstrap(object):
                 os.environ["SHOTGUN_DESKTOP_CURRENT_USER"]
             )
 
+            # If we are using an older version of desktop pre 2.5.4, then
+            # it will try to import PySide which won't be available.
+            # To keep things working, we will import PySide 2 as if it were PySide.
+            self._add_pyside2_as_pyside()
+
             # Prepare the manager based on everything the bootstrap manager expected of us.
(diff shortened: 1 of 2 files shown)


python/utils/bootstrap_utilities.py
```diff
@@ -199,6 +217,34 @@ def _progress_callback(self, value, msg):
...
+        try:
+            import PySide
+        except ImportError:
+            try:
+                import PyQt4
+            except ImportError:
+                # PySide and PyQt4 don't exist so we need to patch the sys.modules to
+                # add PySide2 as if it was PySide. This is so if we are using a
+                # pre v2.5.3 release of tk-desktop it will be able to import QT.
+                from sgtk.util.qt_importer import QtImporter
```
pscadding, line 234: I guess this shouldn't really be used, since it's not a public method.

python/utils/bootstrap_utilities.py
```diff
@@ -148,6 +148,11 @@ def start_engine(self):
                 os.environ["SHOTGUN_DESKTOP_CURRENT_USER"]
```
pscadding, line 148: This is actually for Line 81:
Does this need porting, I would have assumed it would have been covered in the Desktop python 3 support?
It's using `unicode`?

python/utils/bootstrap_utilities.py
```diff
@@ -212,6 +217,35 @@ def _progress_callback(self, value, msg):
...
+                # PySide and PyQt4 don't exist so we need to patch the sys.modules to
+                # add PySide2 as if it was PySide. This is so if we are using a
+                # pre v2.5.3 release of tk-desktop it will be able to import QT.
+                from sgtk.util.qt_importer import QtImporter
+
+                # Importing PySide2 and storing it in sys.modules causes "from PySide import QtCore" not to work.
+                # It import the unpatched version, so we import an empty python file to use as the package.
+                import PySide_patch
+
+                PySide_patch.__name__ = "PySide"
```
jfboismenu, line 240: Could these lines be moved inside PySide_patch?

python/utils/bootstrap_utilities.py
jfboismenu: I don't understand the question. What is the problem exactly?

jfboismenu: The serialize/unserialize user method is python2/3 compliant.

pscadding: ugh, sorry I didn't read it properly, I just saw hound complaining about unicode earlier and didn't spot the check right next to it for Python 2.

PR #113, opened 2020-05-27
SG-17639 Fixes styling issues around docking specifically on Mac.

pscadding:
Sorts the arrow not always being placed in the correct position.
Sorts the issue with the masking and transparency not working. Now looks like this:
Item-0_and_Item-0_and_Item-0_and_Item-0_and_Item-0_and_Item-0


python/tk_desktop/systray.py
```diff
@@ -211,7 +211,7 @@ def _pin_to_menu(self, animated=True):
 
             # setup the animation to shrink the window to the systray
             # parent the anim to self to keep it from being garbage collected
-            anim = QtCore.QPropertyAnimation(self, "geometry", self)
+            anim = QtCore.QPropertyAnimation(self, b"geometry", self)
```
pscadding, line 214: I know you have a fix for this already @jfboismenu, this is here just so that I can test my changes. I'll handle the merge conflicts when I come to merge.

python/tk_desktop/ui/desktop_window.py
```diff
@@ -17,7 +17,7 @@ def setupUi(self, DesktopWindow):
         icon.addPixmap(QtGui.QPixmap(":/tk-desktop/default_systray_icon.png"), QtGui.QIcon.Normal, QtGui.QIcon.Off)
         DesktopWindow.setWindowIcon(icon)
         DesktopWindow.setDockNestingEnabled(False)
-        DesktopWindow.setUnifiedTitleAndToolBarOnMac(True)
+        DesktopWindow.setUnifiedTitleAndToolBarOnMac(False)
```
pscadding, line 20: This is what was causing the transparency issues on the docking on Mac.
Setting it to false doesn't appear to make an difference when running in PySide, but it does fix the PySide2.
Qt dock advise against using this in Qt5.
QT 5: https://doc.qt.io/qt-5/qmainwindow.html#unifiedTitleAndToolBarOnMac-prop
QT 4: https://doc.qt.io/archives/qt-4.8/qmainwindow.html#unifiedTitleAndToolBarOnMac-prop

python/tk_desktop/systray.py
```diff
@@ -443,7 +447,6 @@ def _set_window_mask(self):
 
             if side == self.DOCK_TOP:
                 anchor_pixmap = self.__top_anchor
-                anchor_center = anchor_pixmap.rect().center()
```
pscadding, line 446: This wasn't used so I removed it.

PR #114, opened 2020-05-27
SG-17556: pin menu on top

jfboismenu:
This branch does two things.
First, it fixes an issue where the desktop wouldn't reappear after being dismissed in pinned mode.
The second is that is packages new versions of `osutils` for macOS. After careful testing, only the macOS build of Desktop with PySide2 still needs some help getting desktop properly focussed and on top of other dialogs. The Windows and Linux version do not need them, so no new libraries were packaged.


python/tk_desktop/desktop_window.py
```diff
@@ -56,10 +56,7 @@
 
 from .notifications import NotificationsManager, FirstLaunchNotification
 
-try:
-    from .extensions import osutils
-except Exception:
-    osutils = None
+from .extensions import osutils
```
jfboismenu, line 59: I've rewritten extensions so it looks like it always is there and we handle internally when it is missing. There's less of a chance of making a mistake by not testing for it now.

python/tk_desktop/systray.py
```diff
@@ -332,25 +325,21 @@ def activate(self):
...
-                osutils.activate_application()
+            osutils.activate_application()
 
     def toggle_activate(self):
         """
         Toggles visibility when systray icon is clicked.
         """
-        active = self.isActiveWindow()
-        if active:
+        if self.isVisible():
```
jfboismenu, line 339: This is the actual fix for the pinned dialog. In PySide2, `isActiveWindow` always returns true for some reason. Using isVisible makes sense and works for both PySide and PySide2, so we're good now for all versions of Desktop.

python/tk_desktop/extensions/__init__.py
```diff
@@ -10,9 +10,58 @@
...
+            from .linux_python2_qt4 import osutils
+
+            use_mocked_osutils = False
+        except Exception as e:
+            logger.warning("Could not load osutils: %s", e, exc_info=True)
+elif sys.platform == "darwin":
+    # On macOS, the osutils are required to make the app move to the foreground
+    # or background in certain cases.
+    try:
+        if sys.version_info[0] == 2:
```
pscadding, line 38: Why on Mac is the version dependent on python version, but on Linux its on PySide version?

python/tk_desktop/systray.py
```diff
@@ -163,14 +157,13 @@ def state(self, value):
...
-            if osutils is not None:
-                osutils.make_app_background()
+            osutils.make_app_background()
 
         elif self.__state == self.STATE_WINDOWED:
             self._set_window_mask()
             self.setWindowFlags(self.windowFlags() & ~QtCore.Qt.FramelessWindowHint)
             self.show()
-            if osutils is None:
+            if osutils.is_mocked:
```
pscadding, line 166: Does this work if it is not mocked?
I don't see `osutils.is_mocked = False` being patched in, is it natively there?

pscadding: Also further down, (ln 326 -328), we are raising and then running `make_app_foreground`.
Is there a need to make this an if else? Can we just do both and remove the need for the `is_mocked`?

jfboismenu: Good point! I wasn't thinking enought about the code and just trying to replicate the flow.

python/tk_desktop/extensions/__init__.py
jfboismenu: See comments above (which I've just added)

PR #117, opened 2020-05-29
SG-17640 Macos item highlighting fix

pscadding:
This fixes an issue where when entering a project in SG Desktop on Mac, the proxy Python process bootstrapping the project config, would steal focus when it spawned its app.

The fix here is to push the spawn proxy app to the background, and this will result in the previously active app becoming active again. So in most cases this will mean Shotgun Desktop becomes active again, however if for example you clicked on a project and then went to chrome, it would not return focus to chrome rather than stealing it, which seems like the right solution.

The down side to this fix is that the project config as well as the site config will need to be updated to use the latest version of `tk-desktop`.

Potentially this releasing of focus could happen in the bootstrap_utilities and thus not require the project to update tk-desktop. However, I couldn't seem to import `osutils` on Mac without @jfboismenu's update: 16a76b8b5b6b61c365e362c924f9ae08296968a2.


pscadding:
Actually I've realised this fix causes Desktop to hang if you use this version with an old version of Shotgun Desktop, it needs further looking into.

jfboismenu:
Alright, I'll wait to do the code review then.

pscadding:
OK this is ready for review again. The fix only works on Mac Python 3 PySide2.

I looked into implementing a command back to the main project to restore focus, and whilst I could restore focus, I couldn't seem to record whether the window had focus originally so, I had no way of knowing if I should restore focus or not.

The approach here whilst limited to Python3, doesn't need to communicate to the main process, and restores focus to what ever window previously had focus whether it was Shotgun or not, which seems like the better behaviour.

python/tk_desktop/desktop_engine_project_implementation.py
```diff
@@ -257,6 +289,16 @@ def start_app(self):
...
 
         if self._engine.has_ui:
+
+            if sgtk.util.is_macos():
+                # If we are on Mac with PySide2, then starting a QApplication even with no Windows
+                # will steal focus from any currently focused application.
+                # So we need to deactivate the QApplication after it is launched.
+                if hasattr(app, "applicationStateChanged"):
+                    # Qt 5 has a signal that is triggered when the application state changes which
+                    # we can use to track when teh app takes focus.
```
jfboismenu, line 299:
```suggestion
                    # we can use to track when the app takes focus.
```

PR #119, opened 2020-06-09
SG-17782 Adjusts some of the spacing between recent items.

pscadding:
The recent items list was cropping some of the text on the items depending on the size of the window when you opened the project. This is it in an extreme state where the window was really small when clicking on the project, and then expanded after loading.
Shotgun

This fix does the following:
- Makes the buttons stretch out as the window stretches rather than keeping them fixed.
- Makes the text stay aligned to the top, so that it doesn't display at varying heights when the text for some buttons goes onto multiple lines.
- Reduces the spacing between the buttons slightly (by 1px) so that the impact of having a smaller window is slightly reduced on the cropping.

In summary the cropping of text still happens, but it is less severe, and expanding the window allows the items have enough room to show all their text.
I tested in the PySide2 and PySide builds of desktop.
A video of this in action can be found on the internal ticket.

Shotgun

### DIFF
--- a/python/tk_desktop/command_panel/recent_button.py
+++ b/python/tk_desktop/command_panel/recent_button.py
@@ -20,7 +20,7 @@ class RecentButton(QtGui.QPushButton):
     a single action, unlike the CommandButton.
     """
 
-    MARGIN = 5
+    MARGIN = 2
     SPACING = 5
     SIZER_LABEL = None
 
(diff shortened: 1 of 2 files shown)


python/tk_desktop/command_panel/recent_button.py
```diff
@@ -97,16 +100,3 @@ def command_name(self):
         Name of the command.
         """
         return self._command_name
-
-    def sizeHint(self):
```
pscadding, line 101: This doesn't get reevaluated when the window size changes, so basing the width off the parent widget size meant the size got stuck and didn't grow when expanding the window.
Having the items expand as the window expands removes the need for this method to be implented.

python/tk_desktop/command_panel/recent_list.py
```diff
@@ -22,7 +22,7 @@ class RecentList(BaseIconList):
 
     def __init__(self, parent):
         super(RecentList, self).__init__(parent, QtGui.QHBoxLayout())
-        self._layout.addStretch(1)
```
pscadding, line 25: This line was what was stopping the buttons from stretching out as the window expanded.

python/tk_desktop/command_panel/recent_list.py
```diff
@@ -70,18 +70,15 @@ def add_command(self, command_name, button_name, icon, tooltip, timestamp):
         button.command_triggered.connect(self.command_triggered)
         self._layout.insertWidget(insert_pos, button)
 
-        # If there are now more recents than we should have in the gui,
-        # drop the last one. - 1 applies here because the last item is
-        # the stretcher, which isn't a button.
-        if (self._layout.count() - 1) > MAX_RECENTS:
+        # If there are now more recents than we should have in the GUI,
+        # drop the last one.
+        if self._layout.count() > MAX_RECENTS:
```
pscadding, line 75: Since I removed the stretcher, we no longer needed the -1

pscadding:
I hadn't tried on Windows, but I have now. All looks good in both pinned and unpinned on both PySide and PySide2. No horizontal scroll bars, and it looks visually the same as on Mac.

pscadding:
I've requested a rereview as I've change some stuff. It now adds placeholder stretchers to start with, which get replaced by the recent items. Previously in my fix if you had one recent item it would just be centered. The stretchers now mean the buttons don't jump around when you add recent items.

PR #120, opened 2020-06-11
SG-17861 Fixes an error that gets thrown when leaving a project on OSX.

pscadding:
It seems it errors if the connection gets closed mid loop, then the `select.select` call will error if the listen object is closed.
I'm not sure if this is the right way of going about handling this.

### DIFF
--- a/python/tk_desktop/rpc.py
+++ b/python/tk_desktop/rpc.py
@@ -256,7 +256,7 @@ class RPCServerThread(threading.Thread):
     def run(self):
         """
         Run the thread, accepting connections and then listening on them until
-        they are closed.  Each message is a call into the function table.
+        they are closed. Each message is a call into the function table.
         """
         logger.debug("server listening on '%s'", self.pipe)
         while self._is_closed is False:
@@ -278,10 +278,18 @@ class RPCServerThread(threading.Thread):
                         raise
                     ready = False
             else:
-                # can use select on osx and linux
-                (rd, _, _) = select.select(
-                    [self.server._listener._socket], [], [], self.LISTEN_TIMEOUT
-                )
+                # Can use select on OSX and Linux.
+                try:
+                    logger.info("_socket %s" % self.server._listener._socket)
+                    (rd, _, _) = select.select(
+                        [self.server._listener._socket], [], [], self.LISTEN_TIMEOUT
+                    )
+                except select.error as e:
+                    if "Bad file descriptor" in e.args:
+                        # The socket was likely closed during this iteration of the loop
+                        # so we should break out of the loop.
+                        break
+                    raise
                 ready = len(rd) > 0
 
             if not ready:


python/tk_desktop/rpc.py
```diff
@@ -278,10 +278,17 @@ def run(self):
...
-                (rd, _, _) = select.select(
-                    [self.server._listener._socket], [], [], self.LISTEN_TIMEOUT
-                )
+                # Can use select on OSX and Linux.
+                try:
+                    (rd, _, _) = select.select(
+                        [self.server._listener._socket], [], [], self.LISTEN_TIMEOUT
+                    )
+                except select.error as e:
+                    if "Bad file descriptor" in e.args:
```
pscadding, line 287: Is there a better way of catching this?

pscadding: We could in theory wait a short while and check the `self._is_closed` variable before deciding to break or raise.
I noticed that sometimes if I printed the `self._is_closed` in the exception block it would be false and sometimes true. I presume this is because it failed due to it being in the process of closing, but hadn't actually finished so the flag wasn't set.

pscadding:
I looks like there are some other errors thrown at other stages such as if you press back whilst loading a project or if you close desktop whilst loading a project. I'll make those part of this PR so actually hold off on the review for now.

tests/test_rpc.py
```diff
@@ -112,6 +112,25 @@ def long_call(self, close_proxy, return_value=None):
         return return_value
 
 
+class CaptureErrorServer(RPCServerThread):
+    """
+    This is the RPCServerThread but it captures any errors
```
jfboismenu, line 117: Ah, I like it! Smart.

pscadding: yeah I like it as well but as per our discussion I'll remove it, as we are not going to go and fix the issues it is raising right now in the tests.

It did actually however prove that my fix worked.

pscadding: OK I've actually left it in and commented out the raising. How do you feel about that?
Would you prefer me to remove it entirely. I feel by leaving it in there we are drawing attention to the fact that we are not capturing the errors from the server and providing a solution that could be enabled when we are ready to deal with the errors.

another approach I guess would be to remove the class but just leave a TODO with the link to the SO answer that gave me the solution.

jfboismenu: Well, if you have inspired the code from something you saw on StackOverflow, you are legally required to include a link. All posts on StackOverflow are done under the [Creative Commons](https://creativecommons.org/licenses/by-sa/4.0/) license. It requires that any piece of code you reuse or adapt is given attribution by providing a link to the required post.

I also just learned that you need to mention if the code was modified or as is. I've generally been using the following formulations: "This code has been lifted" vs "This code has been inspired by".

jfboismenu: I'd rather put a comment in the code itself with a FIXME saying that a excetion can be raised from here, so the next person who needs to fix this can do it can find the place to fix right away.

pscadding: AH OK I didn't know about the inspired by bit, but I did have the link in there.

I have a TODO which I could rename to FIXME, are you saying just have a comment/FIXME with the link and remove the class entirely rather than just commenting out the raise bit in the class?

jfboismenu: From my understand the class is not used. Instead of leaving a test class with some magic code in it, let's document the actual class with its shortcoming. Does that make sense?

pscadding:
I've changed the approach again. Francis was getting a different error than originally reported when testing this. And when I went back to test it I was getting this different error as well, and I couldn't reproduce the originally reported error.
I think it comes down to luck/speed, and depending when the server gets closed, you get a different error. 
So I thought probably the best thing to do is catch all errors, and then check the `self._is_closed` property, however its possible this hasn't yet been set, so I've included a small sleep to give it a chance, and then if it is still not set, raise the error.

What do you think?

PR #123, opened 2020-06-30
SG-17970 Fixes recent items not showing.

pscadding:
Recent items were not being stored when in Python 3 due to the way we were trying to pickle the datetime objects before saving them with `QSettings`.

I looked into applying the fix at the shotgunutils side, and whilst it was possible there were other complications and things to consider. In the end it was simpler to store string representations of the datetime instead.

This code handles loading both datetime object and string dates which it then converts back into datetime objects for use during run time.

You will get an error if you downgrade desktop as the old desktop will not know how to handle the string format. However this error is largely silent, it only appears to show if you launched from a terminal, and the rest of the recents items that weren't stored with a string format do show up.

### DIFF
--- a/python/tk_desktop/command_panel/command_panel.py
+++ b/python/tk_desktop/command_panel/command_panel.py
@@ -8,13 +8,15 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-import datetime
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


pscadding:
We're avoiding pickling the datetime object as this causes issues in Python 3, and cross python reading.

As far as I can see there, this change doesn't make Python 2.7 lose it's recents, and whilst it was a few weeks ago now, I believe I tested that it didn't. What makes you think it will?

jfboismenu:
Sorry! I got it backward, it's all good!

PR #127, opened 2020-07-30
SG-18612 fix for no keyboard input on mac

pscadding:
Fixes an issue that was introduced with the fix for the project `QApplication` stealing focus on Mac.

That fix introduced an an issue where the keyboard input couldn't be registered with any of the launched apps. This was because the `QApplication` was permanently in the background.

The fix is to now move it to the foreground when we run an app.

### DIFF
--- a/python/tk_desktop/desktop_engine_project_implementation.py
+++ b/python/tk_desktop/desktop_engine_project_implementation.py
@@ -189,7 +189,7 @@ class DesktopEngineProjectImplementation(object):
             # will steal focus from any currently focused application.
             # We set the application to use `LSUIElement=1` so that it stays in the background in start_app().
             # However we need to bring it to the foreground when we want to run an app.
-            # This solution for bring it back to the foreground is a modification of the example here:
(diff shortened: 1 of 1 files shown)


python/tk_desktop/desktop_engine_project_implementation.py
```diff
@@ -183,6 +183,29 @@ def destroy_engine(self):
 
     def _trigger_callback(self, namespace, command, *args, **kwargs):
         callback = self.__callback_map.get((namespace, command))
+
+        if sgtk.util.is_macos():
+            # If we are on Mac with PySide2, then starting a QApplication even with no Windows
+            # will steal focus from any currently focused application.
+            # We set the application to use `LSUIElement=1` so that it stays in the background in start_app().
+            # However we need to bring it to the foreground when we want to run an app.
+            # This solution for bring it back to the foreground is a modification of the example here:
```
pscadding, line 192:
```suggestion
            # This solution for bringing it back to the foreground is a modification of the example here:
```

PR #128, opened 2020-08-19
SG-18905: Site switching fix

jfboismenu:
When another site tries to connect to browser integration in Shotgun Desktop, the site switching dialog did not appear in Python 3. This was because the site switching code had been disabled back when the browser integration was not Python 3 compatible, but we forgot to turn it back on. So now it is.

I also noticed an issue where the site switching dialog would not be focussed on macOS. Turns out we were importing a module from a directory without a `__init__.py` on macOS, so I've added it. On Linux the the `__init__.py` file was present and on Windows there is no extension module to load.

### DIFF
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -944,12 +944,10 @@ class DesktopWindow(SystrayWindow):
 
             dialog = BrowserIntegrationUserSwitchDialog(msg, self)
 
-            # The following applies to macOS only and has no side-effect on other plaforms.
-            # If the dialog is pinned, it means it is also in background. We'll bring the app to the foreground
-            # so keyboard focus is granted automatically to the BrowserIntegrationUserSwitchDialog instead
-            # of being unfocussed.
-            if self.is_pinned():
-                osutils.make_app_foreground()
+            # The following applies to macOS only and has no side-effect on other platforms.
+            # Making the app the foreground app on macOS ensure it gets the focus. Other platforms
+            # just work without this call.
+            osutils.make_app_foreground()
 
             dialog.exec_()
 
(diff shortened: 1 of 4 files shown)


python/tk_desktop/desktop_window.py
```diff
@@ -947,12 +944,10 @@ def _on_different_user(self, site, user_id):
 
             dialog = BrowserIntegrationUserSwitchDialog(msg, self)
 
-            # The following applies to macOS only and has no side-effect on other plaforms.
-            # If the dialog is pinned, it means it is also in background. We'll bring the app to the foreground
-            # so keyboard focus is granted automatically to the BrowserIntegrationUserSwitchDialog instead
-            # of being unfocussed.
-            if self.is_pinned():
-                osutils.make_app_foreground()
+            # The following applies to macOS only and has no side-effect on other platforms.
```
jfboismenu, line 947: Seems like in PySide2 we need to always make this call now. As for making this call in PySide1, this code has no noticeable side-effect when the dialog is not pinned, so I've simplified the code.

python/tk_desktop/extensions/darwin_python2/__init__.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2020 Shotgun Software Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
```
jfboismenu, line 5: There are the missing `__init__.py` that prevented the `osutils.so` file from being loaded in the `darwin_python2` and `darwin_python3` folders.

python/tk_desktop/desktop_window.py
```diff
@@ -277,8 +276,7 @@ def __init__(self, console, parent=None):
 
         # This framework is not compatible with Python 3 at the moment
```
eshokrgozar, line 277: We should probably remove this comment now too

python/tk_desktop/desktop_window.py
```diff
@@ -367,10 +365,9 @@ def __init__(self, console, parent=None):
         )
 
         # This framework is not compatible with Python 3 at the moment.
```
eshokrgozar, line 367: and this one

PR #132, opened 2020-11-24
SS-18239: undocked by default

jfboismenu:
The Shotgun Desktop will now be displayed unpinned by default the first time it is launched. This will reduce the confusion of first time users who do not understand where the app goes after the splash screen disappears.

Also, when a user clicks on the X or red button of the dialog, a pop-up will be shown the very first time letting the user know the app is now in the tray.

Finally, unhandled exceptions are now going to be logged. (FINALLY!!!!)

UPDATE after some QA tests:

I've rewired the chain of quitting operations in the Desktop. The Quit option in the menu instead calls quit on the QApplication and the handle_quit_action has been turned into a slot for the QApplication.aboutToQuit signal. This was necessary to handle macOS quirkyness when hitting CMD-Q properly. See the comments in closeEvent.

Note that this only works in the Qt5 build of Desktop. Qt4 build doesn't seem to support Catalina's notifcation/permission system.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -11,6 +11,7 @@
 import sys
 
 import sgtk
+import traceback
 from sgtk.platform import Engine
 
 
@@ -79,6 +80,21 @@ class DesktopEngine(Engine):
                 interface_type = "project"
 
         self._is_site_engine = interface_type == "site"
+
+        previous_except_hook = sys.excepthook
+
+        def unhandled_exception_handler(exc_type, exc_value, exc_traceback):
+            engine = sgtk.platform.current_engine()
+            if engine:
+                self.logger.error(
+                    "".join(
+                        traceback.format_exception(exc_type, exc_value, exc_traceback)
+                    )
+                )
+            if previous_except_hook:
+                previous_except_hook(exc_type, exc_value, exc_traceback)
+
+        sys.excepthook = unhandled_exception_handler
 
         # Import our python library
         #
--- a/engine.py
+++ b/engine.py
@@ -14,6 +14,25 @@ import sgtk
 import traceback
 from sgtk.platform import Engine
 
+previous_except_hook = sys.excepthook
+
+
+def unhandled_exception_handler(exc_type, exc_value, exc_traceback):
+    # Calls any previous exception handler. By default, this will invoke the
+    # console error handler.
+    if previous_except_hook:
+        previous_except_hook(exc_type, exc_value, exc_traceback)
+
+    engine = sgtk.platform.current_engine()
+    # As long as we have an engine, we can send errors to its logger.
+    if engine:
+        engine.logger.error(
+            "".join(traceback.format_exception(exc_type, exc_value, exc_traceback))
+        )
+
+
+sys.excepthook = unhandled_exception_handler
+
 
 class DesktopEngine(Engine):
     def __init__(self, tk, *args, **kwargs):
@@ -80,21 +99,6 @@ class DesktopEngine(Engine):
                 interface_type = "project"
 
         self._is_site_engine = interface_type == "site"
-
-        previous_except_hook = sys.excepthook
-
-        def unhandled_exception_handler(exc_type, exc_value, exc_traceback):
-            engine = sgtk.platform.current_engine()
-            if engine:
-                self.logger.error(
-                    "".join(
-                        traceback.format_exception(exc_type, exc_value, exc_traceback)
-                    )
-                )
-            if previous_except_hook:
-                previous_except_hook(exc_type, exc_value, exc_traceback)
-
-        sys.excepthook = unhandled_exception_handler
 
         # Import our python library
         #
--- a/engine.py
+++ b/engine.py
@@ -13,22 +13,41 @@ import sys
 import sgtk
 import traceback
 from sgtk.platform import Engine
+from sgtk import LogManager
+
+logger = LogManager.get_logger(__file__)
 
 previous_except_hook = sys.excepthook
 
 
 def unhandled_exception_handler(exc_type, exc_value, exc_traceback):
+    """
+    Unhandled exception handler.
+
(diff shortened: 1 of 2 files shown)


engine.py
```diff
@@ -11,8 +11,28 @@
...
+    # console error handler.
+    if previous_except_hook:
+        previous_except_hook(exc_type, exc_value, exc_traceback)
+
+    engine = sgtk.platform.current_engine()
+    # As long as we have an engine, we can send errors to its logger.
+    if engine:
+        engine.logger.error(
+            "".join(traceback.format_exception(exc_type, exc_value, exc_traceback))
+        )
```
pscadding, line 31: I'm not sure on when events where this method is called, but this will catch any errors raised after the engine has started. What if an app errors during initialisation, does it get logged then by some other code?

python/tk_desktop/desktop_window.py
```diff
@@ -469,7 +473,9 @@ def _load_settings(self):
         self.set_on_top(self._settings_manager.retrieve("on_top", False))
 
         # always start pinned and hidden
```
pscadding, line 475:
```suggestion
        # always start unpinned and shown
```
Should the comment be this instead?

pscadding (approved):
Small comment correction needed but looks good, great to see this finally improved!!

engine.py
jfboismenu: Ah, good point. If the engine hasn't started yet, we're going to miss it. Maybe I should log directly to `sgtk.LogManager.get_logger(__file__)` instead! I'll try it out. Thanks!

pscadding: Nice!

python/tk_desktop/desktop_window.py
```diff
@@ -600,6 +605,64 @@ def _register_apps_tab(self):
     def contextMenuEvent(self, event):
         self.user_menu.exec_(event.globalPos())
 
+    def closeEvent(self, event):
+        # When closing the app programatically via self.close(),
+        # the event is not spontaenous and we know the state of self._is_quitting
```
pscadding, line 610: A small thing here, take it or leave it, but I had to look this up, because I wasn't sure if you were just calling the event spontaneous or if an event actually had a spontaneous state/parameter.
In the case of this I read that it is an event that was triggered from outside of the application.

python/tk_desktop/desktop_window.py
```diff
@@ -600,6 +605,64 @@ def _register_apps_tab(self):
...
+        # is correct.
+
+        # When CMD-Q is hit or the close button is pressed however, the dialog
+        # first receives a closeEvent and THEN receives the aboutToQuit signal,
+        # which will call the close event. When this happens, the event is marked
+        # as spontaneous, but we can't tell what actually happened so we postpone
+        # the execution of it by making sure it gets called again after the event
+        # queue is emptied.
```
pscadding, line 618: > making sure it gets called again after the event queue is emptied.

I'm not seeing anything here that checks if the event queue is empty. Basically I'm not sure what piece of logic this line is referring to. Is it that the Timer is giving the process enough time to complete all the events?

python/tk_desktop/desktop_window.py
```diff
@@ -600,6 +605,64 @@ def _register_apps_tab(self):
...
+        # When the timer has expired, we will know if the spontaneous event had
+        # been a simple close of the dialog OR if the app was closed.
+        # If is_quitting isn't set by that point, it's because the user pressed
+        # the close button. If it is, then we know the app is closing and we know
+        # how to deal with that.
+        if event.spontaneous() is False or self._is_quitting:
+            self._handle_close_event()
```
pscadding, line 625: Why would we call _handle_close_event if _is_quitting true? It doesn't do anything?

python/tk_desktop/desktop_window.py
```diff
@@ -600,6 +605,64 @@ def _register_apps_tab(self):
...
+        else:
+            QtCore.QTimer.singleShot(0, self._handle_close_event)
+
+    def _handle_close_event(self):
+        if self._is_quitting:
```
pscadding, line 630: So this method, if we assume for a sec that _is_quitting is false, will just handle the minimize of the app and not the shutting down?

pscadding: Maybe it should have a doc string?

python/tk_desktop/desktop_window.py
```diff
@@ -698,7 +761,8 @@ def handle_systray_state_changed(self, state):
             self.ui.actionPin_to_Menu.setText("Pin to Menu")
         self._settings_manager.store("dialog_pinned", self.state)
 
-    def handle_quit_action(self):
+    def _about_to_quit(self):
+        self._is_quitting = True
```
pscadding, line 765: Should this method have a doc string?

pscadding (commented):
Wow this is a real head scratcher. I've reread the comment several times and I think I've got it but I'm still not certain.

Essentially what we are trying to do is understand if Desktop is being instructed to actually shut down or the user has pressed the close button and we want to "minimize" to the task tray.

So when the user is pressing CMD+Q we are actually wanting to shut the application down, but when we are pressing the close/cross button we are not wanting to shut the application down, just go to a minimized state?

python/tk_desktop/desktop_window.py
jfboismenu: Yeah, I did a poor job here, I'll clarify the comments

python/tk_desktop/desktop_window.py
jfboismenu: Yeah, I don't understand how this works anymore... let me go back to test it and improve the comments. Thanks for keeping me honest! 😄

pscadding (approved):
Yeah that looks great, It's not a simple thing to explain but that's much clearer. Thanks for doing that!

PR #134, opened 2020-12-15
Fix for restoring window position error

pscadding:
If the previous window position was stored as a PySide object it can sometimes end up getting through the `settings_manager.retrieve` and passed to PySide2, on which it then fails. This now just gracefully gives up on restoring the window position in this case.

Still needs testing, but getting this scenario is difficult and its not entirely clear how to replicate.


eshokrgozar (approved):
I don't quite understand the situation leading up to this but the fix makes sense to me.

pscadding:
I'm not entirely sure why this is happening either. It's hard to replicate. Usually if you are using PySide2 and you try to restore a saved PySide object from the settings manager, the settings manager just passes back `None`. However in some cases it manages to pass back a PySide object which would then have caused the error. We're not sure what you need to do however to get the environment in a state where it does this.

PR #135, opened 2021-01-21
Fixes broken software credit links, and adds back in missing ones.

pscadding:


mathurf (approved):
LGTM!

:shipit:

python/tk_desktop/licenses.html
```diff
@@ -1144,20 +1144,19 @@
     Shotgun Desktop can be found here:
     <ul>
         <li><a href="https://github.com/shotgunsoftware/python-api/blob/master/software_credits">Shotgun Python API</a></li>
-        <li><a href="https://github.com/shotgunsoftware/tk-alias/blob/master/software_credits">tk-alias</a></li>
```
jfboismenu, line 1147: Given how repetitive these links are, would it make sense to create a list of repos and write a loop that generates all links? This way we would have a short list of repos we could quickly look at and figure more easily if something is missing?

jfboismenu (commented):
I know this has been merged, but I'd like to provide food for thought for the next time we update this list.

python/tk_desktop/licenses.html
jfboismenu: Oh, this isn't Python, right! We'd have to do it with a JS script at the top of the file, so maybe not, since I'm not sure Qt4 would even process it. :(

pscadding: well we could I suppose write a test that at least checks all the links are present and point to valid urls. I've not tried that before, but I presume we could parse the html and extract the links and check we don't get a 404?

jfboismenu: Yeah, not a bad idea actually!

pscadding: It wouldn't generate the list, but It would fit in the existing framework and process, and then notify us if something went wrong.

jfboismenu: You can add it now if you feel like it, but there's no urgency. I'm sure Francis has clicked on all of them during QA. 😄

PR #136, opened 2021-01-21
Testing licence urls

pscadding:
Adds tests that check that all the URLs in the licences.html is a valid link, and checks that certain expected software credit link exist.

### DIFF
--- a/tests/test_about_box.py
+++ b/tests/test_about_box.py
@@ -1,0 +1,112 @@
+# Copyright (c) 2020 Shotgun Software Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import os
+import pytest
+from six.moves.html_parser import HTMLParser
+from six.moves.urllib import request
+
+
+@pytest.fixture
+def license_file():
+    """
+    Gets the path to the licence html file.
+    """
+    license_file = os.path.abspath(
+        os.path.join(
+            os.path.dirname(__file__), "..", "python", "tk_desktop", "licenses.html"
+        )
+    )
+    assert os.path.isfile(license_file)
+    return license_file
+
+
+@pytest.fixture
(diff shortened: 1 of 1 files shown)


tests/test_about_box.py
```diff
@@ -0,0 +1,112 @@
...
+
+def test_3rd_party_links(licence_file_links):
+    """
+    Check all found urls are valid and can accessed.
+    """
+    for url in licence_file_links:
+        try:
+            request.urlopen(url)
+        except Exception as e:
+            raise pytest.fail("Failed to open {0}, error: {1}".format(url, e))
```
pscadding, line 69: I could have just let it fail and report the error, but then it wouldn't have logged which url failed.

tests/test_about_box.py
```diff
@@ -0,0 +1,112 @@
...
+            # Only parse the 'anchor' tag.
+            if tag == "a":
+                # Check the list of defined attributes.
+                for name, value in attrs:
+                    # If href is defined, store the URL
+                    if name == "href":
+                        self.tags.append(value)
+
+    # Open the licence file and read the parse the contents.
+    f = open(license_file, "r")
```
jfboismenu, line 52: Please use `with open(...) as fd`. Think of the children.

tests/test_about_box.py
jfboismenu: Smart

tests/test_about_box.py
```diff
@@ -0,0 +1,112 @@
...
+        (
+            "https://github.com/shotgunsoftware/tk-framework-desktopstartup/blob/master/python/server/software_credits"
+        ),
+        (
+            "https://github.com/shotgunsoftware/tk-framework-lmv/blob/master/software_credits"
+        ),
+        ("https://github.com/shotgunsoftware/tk-houdini/blob/master/software_credits"),
+        (
+            "https://github.com/shotgunsoftware/tk-multi-launchapp/blob/master/software_credits"
+        ),
```
jfboismenu, line 96: Why the parantheses? Black did this?

jfboismenu (approved):
Looks alright, but I do have an issue with the formatting of the URL list at the end of the file?

tests/test_about_box.py
jfboismenu: If we could remove them I would appreciate it. This makes the list harder to read as it zigs and zags.

pscadding: ah, its because I copied it from the Python console tests I wrote, and they had multiple parameters, I think I just assumed it was part of the syntax, but you're right, they don't need to be there.

PR #144, opened 2022-01-26
SG-24622 Update SG Desktop to use pyside 5 15

000paradox000:
This PR fix an issue when switching and opening last visited project.

Additionally it polish PEP-8 in the files where changes were applied.

### DIFF
--- a/python/tk_desktop/project_model.py
+++ b/python/tk_desktop/project_model.py
@@ -209,11 +209,15 @@ class SgProjectModelProxy(QtGui.QSortFilterProxyModel):
         """
         if not hasattr(self, "_ids_in_order"):
             """
-            The cache probably is not ready yet. With Qt5.15 sometimes we are called by setSourceModel so for sure the 
-            cache is not  ready yet. The best value to return here is TRUE which is the default in Qt. Hopefully we 
-            will be called again once the cache is up-to-date and ready to be use.
+            The cache probably is not ready yet. With Qt5.15 sometimes we are 
+            called by setSourceModel so for sure the cache is not  ready yet. 
+            
+            The best value to return here is TRUE which is the default in Qt. 
+            Hopefully we will be called again once the cache is up-to-date and 
+            ready to be use.
 
-            Please don't remove the following line otherwise we will have a major showstopper for one of our big clients
+            Please don't remove the following line otherwise we will have a 
+            major showstopper for one of our big clients
             Ref: SG-24622 at Autodesk
             """
             return True
--- a/python/tk_desktop/project_model.py
+++ b/python/tk_desktop/project_model.py
@@ -213,7 +213,7 @@ class SgProjectModelProxy(QtGui.QSortFilterProxyModel):
             cache is not  ready yet. The best value to return here is TRUE which is the default in Qt. Hopefully we 
             will be called again once the cache is up-to-date and ready to be use.
 
-            Please don’t remove the following line otherwise we will have a major showstopper for one of our big clients
+            Please don't remove the following line otherwise we will have a major showstopper for one of our big clients
             Ref: SG-24622 at Autodesk
             """
             return True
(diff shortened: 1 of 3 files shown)


python/tk_desktop/rpc.py
```diff
@@ -422,15 +430,16 @@ def call(self, name, *args, **kwargs):
...
                         raise RuntimeError("client closed while waiting for a response")
                     continue
             except IOError:
-                # An exception is raised on Windows when the connection is closed
-                # during polling instead of simply returning False.
+                # An exception is raised on Windows when the connection
+                # is closed during polling instead of simply returning False.
                 if self._closed:
-                    raise RuntimeError("client closed while waiting for a response")
+                    raise RuntimeError("client closed while waiting for a " "response")
```
geffrak, line 442: weird change.

python/tk_desktop/rpc.py
```diff
@@ -439,7 +448,7 @@ def call(self, name, *args, **kwargs):
         except OSError:
             # On Linux, an exception will be raised here instead.
             if self._closed:
-                raise RuntimeError("client closed while waiting for a response")
+                raise RuntimeError("client closed while waiting for a " "response")
```
geffrak, line 451: weird change

python/tk_desktop/project_model.py
```diff
@@ -207,6 +207,21 @@ def filterAcceptsRow(self, source_row, source_parent):
         """
         QSortFilterProxyModel override to base filtering on our cached values.
         """
+        if not hasattr(self, "_ids_in_order"):
+            """
+            The cache probably is not ready yet. With Qt5.15 sometimes we are
+            called by setSourceModel so for sure the cache is not  ready yet.
+
+            The best value to return here is TRUE which is the default in Qt.
+            Hopefully we will be called again once the cache is up-to-date and
```
geffrak, line 216: How can we validate that?

python/tk_desktop/project_model.py
```diff
@@ -207,6 +207,21 @@ def filterAcceptsRow(self, source_row, source_parent):
...
+            ready to be use.
+
+            Please don't remove the following line otherwise we will have a
```
geffrak, line 219: I don't think this part of the comment is required.

geffrak (approved):

python/tk_desktop/project_model.py
000paradox000: In QT filterAcceptsRow returns True by default according to docs https://doc.qt.io/qt-5/qsortfilterproxymodel.html#filterAcceptsRow

We want to filter only when the cached values are ready.

000paradox000: We confirmed that the method is called again debugging the code with breakpoints.

NorberMV (approved):
LGTM!

python/tk_desktop/rpc.py
```diff
@@ -360,9 +367,10 @@ def close(self):
         logger.debug("server setting flag to stop")
         self.server.close()
         self._is_closed = True
-        if sys.platform == "win32":
-            # The accept call blocks until a client is available.
-            # Unfortunately closing the connection on Windows does not
+
+        if six.PY2 and sys.platform == "win32":
```
geffrak, line 371: @mathurf, do you recall a bug that happened when you were validating SG-14617? This got added at that point and doesn't look natural. It looks like a patch to fix an issue encountered while working on that feature.

mathurf (approved):
I tested:
1.7.2 with Python 2/3,
1.7.1 with Python 2/3,
1.5.9 (python 2 only)
Everything looks good

python/tk_desktop/rpc.py
mathurf: Thanks for the ticket id. I took a look and all issues I mentioned in that ticket are not happening.

geffrak (approved):

mathurf (approved):
tested on Windows, mac and Linux.
Everything is good on my side.

PR #147, opened 2022-08-08
SG-11032 [tk-desktop] doesn't implement the host_info

vazquea:
In order to assert a creation of _host_info attribute for each class or subclass instances, I added the method `__new__` to `DesktopEngine` class.


NorberMV (approved):
LGTM!

PR #149, opened 2022-10-04
SG-27300 Add banner to SG-Desktop with python 2 deprecate notice

000paradox000:
image


NorberMV (approved):
LGTM, this is great !

PR #152, opened 2022-12-09
SG-20937 - Fix Quit shortcut

dsauve-adsk:
This explains why CTRL is translated to Meta ( https://doc.qt.io/qt-5/qkeysequence.html#standard-shortcuts )
This one explains why we did not saw the shortcut in MAC ( https://bugreports.qt.io/browse/QTBUG-49435 )

I added the shortcut to MAC because I think it make sense and it is what we decided but I can remove it if we prefer align ourself on MAC defaults...


000paradox000 (approved):
LGTM

julien-lang (approved):
LGTM

PR #153, opened 2023-01-18
[SHOT-4179] Do not crop Desktop project thumbnails

barbara-darkshot:
When the projects are loaded in SG Desktop, do not resize their thumbnail to have something square but use the letterbox format instead if the thumbnail size does not fit
A new environment variable SGTK_PROJ_THUMB_OLD has been introduced to keep the old behavior if needed


python/tk_desktop/thumb_widget.py
```diff
@@ -34,11 +36,25 @@ def __init__(self, width=120, parent=None):
...
+        if os.getenv("SGTK_PROJ_THUMB_OLD") == "True":
+            # zoom to fit height, then crop to center
+            pixmap = pixmap.scaledToHeight(
+                self.thumb_size, QtCore.Qt.SmoothTransformation
+            )
+            if pixmap.width() > self.thumb_size:
+                extra = pixmap.width() - self.thumb_size
+                pixmap = pixmap.copy(extra / 2, 0, self.thumb_size, self.thumb_size)
+        else:
+            dst_pixmap = QtGui.QPixmap(self.thumb_size, self.thumb_size)
```
julien-lang, line 49: Is it possible to add a comment to describe what is the strategy (scale?) in this else block? 
Like the `zoom to fit height, then crop to center` in the IF block

NorberMV (approved):
LGTM !, Just a comment, should this new 'SGTK_PROJ_THUMB_OLD' environment variable be documented somewhere ?

python/tk_desktop/thumb_widget.py
rob-aitchison: @julien-lang , I believe this was addressed, can we resolve this conversation? Thanks!

julien-lang (approved):
LGTM!

rob-aitchison:
@barbara-darkshot , the documentation PR's are now also created to support this:
https://github.com/shotgunsoftware/developer.shotgunsoftware.com/pull/413
https://github.com/shotgunsoftware/tk-core/pull/871
Thanks!

PR #156, opened 2023-03-17
SG-29738 Bump copyright year

carlos-villavicencio-adsk:
image

### DIFF
--- a/python/tk_desktop/licenses.html
+++ b/python/tk_desktop/licenses.html
@@ -2,7 +2,7 @@
 <body>
 <!--        Autodesk licence header-->
 <div>
-    <p><strong>ShotGrid Desktop &copy; 2021 Autodesk, Inc. All rights reserved.</strong></p>
+    <p><strong>ShotGrid Desktop &copy; 2023 Autodesk, Inc. All rights reserved.</strong></p>
     <p>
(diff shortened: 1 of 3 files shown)


python/tk_desktop/about_screen.py
```diff
@@ -20,11 +22,16 @@ class AboutScreen(QtGui.QDialog):
...
 
+        year = date.today().year
+
         # setup the GUI
         self.ui = about_screen.Ui_AboutScreen()
         self.ui.setupUi(self)
         self.ui.body.setText(body)
         self.ui.licensesButton.clicked.connect(self._show_licenses)
+        self.ui.copyright.setText(
+            f"Copyright ©{year} ShotGrid Software Inc.<br/>All rights reserved."
```
julien-lang, line 33: The dynamic year cannot be set during execution. It must be set during build/compilation/release IMO

carlos-villavicencio-adsk: I see your point and it makes sense. However, I can't see where I can make a dynamic year without affecting the whole packing process. Let's bump it to 2023 for now, to avoid extending this ticket too much.

julien-lang (approved):
LGTM!

PR #164, opened 2023-11-09
SG-31715 Engine Init Hook Cant Run QtguiQmessagebox

NorberMV:
This checks if Qt is already running and if a QApplication has been previously initialized. If not, it creates a Qt app instance. This fixes the RuntimeError that occurs silently in the desktop bootstrap when create any Qt application in the engine_init core hook.


julien-lang (approved):
So simple, I like it :)

PR #169, opened 2023-12-27
Incorrect srgb profile warnings in SG Desktop when launched from the terminal

NorberMV:


julien-lang (approved):
LGTM! Thanks @NorberMV

PR #171, opened 2024-02-06
SG-34462 Test CI with Python 3.11

carlos-villavicencio-adsk:
- [x] Bump copyright year
- [x] Fix PySide6 compatibility issue
- [x] Revert ci-tools ref


python/tk_desktop/command_panel/command_button.py
```diff
@@ -162,7 +162,7 @@ def add_command(self, command_name, menu_name, icon, tooltip, is_menu_default):
         # If there is more than one available item in the menu, show it so the
         # user can pick one.
         if len(self._menu.actions()) > 1:
-            self.setPopupMode(self.MenuButtonPopup)
```
carlos-villavicencio-adsk, line 165: This is the only change related to PySide6. We're using the enumerator class instead of using it as object attribute. This is backwards compatible as CI confirmed.

python/tk_desktop/extensions/darwin_python310/osutils.cpython-310-darwin.so
carlos-villavicencio-adsk, line 1: This file was generated internally and now it's architecture is `universal2` compatible with Apple Silicon and Intel.

python/tk_desktop/command_panel/command_button.py
```diff
@@ -162,7 +162,7 @@ def add_command(self, command_name, menu_name, icon, tooltip, is_menu_default):
...
+            self.setPopupMode(QtGui.QToolButton.MenuButtonPopup)
```
carlos-villavicencio-adsk, line 165: This is the only change related to PySide6. We're using the enumerator class instead of using it as object attribute. This is backwards compatible as CI confirmed.

julien-lang (approved):
LGTM!

PR #173, opened 2024-05-16
SG-35333 Fix broken NoAppsInstalledOverlay layout

julien-lang:
The CustomNonProjectEntity23 entity is long gone and has been replaced by the Software entity since SG version 7.2.

Fixing the behavior by re-aligning the function behavior with the implementation in tk-multi-launchapp: https://github.com/shotgunsoftware/tk-multi-launchapp/blob/v0.13.0/python/tk_multi_launchapp/software_entity_launcher.py#L204

### DIFF
--- a/python/tk_desktop/no_apps_installed_overlay.py
+++ b/python/tk_desktop/no_apps_installed_overlay.py
@@ -172,7 +172,7 @@ class NoAppsInstalledOverlay(QtGui.QWidget):
 
     def _get_sg_software_entities(self, engine, project=None):
         """
-        Retrieve a list of Software entities from Shotgun that
+        Retrieve a list of Software entities from FPTR that
         are active for the current project and user.
 
         If the shotgun connection does not support software entities,
--- a/python/tk_desktop/no_apps_installed_overlay.py
+++ b/python/tk_desktop/no_apps_installed_overlay.py
@@ -270,7 +270,6 @@ class NoAppsInstalledOverlay(QtGui.QWidget):
 
         return sg_softwares
 
-
     def __get_sg_server_version(self, engine):
         """
         Retrieves the shotgun server version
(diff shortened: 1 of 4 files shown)


python/tk_desktop/no_apps_installed_overlay.py
```diff
@@ -266,3 +269,15 @@ def _get_sg_software_entities(self, engine, project=None):
...
+        """
+        Retrieves the shotgun server version
+        from the currently connected Shotgun.
+
+        :returns: Tuple of (major, minor, patch) versions.
+        """
+        sg_major_ver = engine.shotgun.server_info["version"][0]
+        sg_minor_ver = engine.shotgun.server_info["version"][1]
+        sg_patch_ver = engine.shotgun.server_info["version"][2]
+        return sg_major_ver, sg_minor_ver, sg_patch_ver
```
carlos-villavicencio-adsk, line 283: Is it possible to just rely on ?

```suggestion
        return engine.shotgun.server_info["version"][:3]
```

carlos-villavicencio-adsk (approved):
LGTM. Good job!

python/tk_desktop/no_apps_installed_overlay.py
```diff
@@ -172,15 +172,21 @@ def _set_configuration_link(self, engine):
 
     def _get_sg_software_entities(self, engine, project=None):
         """
-        Retrieve Software entities from Shotgun for the specified Project.
+        Retrieve a list of Software entities from Shotgun that
```
staceyoue, line 175:
```suggestion
        Retrieve a list of Software entities from FPTR that
```

staceyoue (approved):
LGTM

python/tk_desktop/no_apps_installed_overlay.py
julien-lang: Picking individual index raise an IndexError if does not exist. While picking an index range returns whatever length is available. So it's not exactly the same behavior.

Besides, I copied/pasted this function from tk-multi-launchapp and I already created a ticket for refactoring this duplication into a framework.

PR #174, opened 2024-05-17
SG-30560 Code Cleanup: Removing Qt4/PySide/PyQt4 compatibility

eduardoChaucaGallegos:
Code Cleanup, so we are dropping the support and compatibility with Qt4.


engine.py
```diff
@@ -220,7 +220,7 @@ def __getattr__(self, name):
                 raise sgtk.TankError(
                     "Looks like you are trying to run an App that uses a QT based UI, however the "
                     "python installation that the Desktop engine is currently using does not seem "
-                    "to contain a valid PySide or PyQt4 install. Either install PySide into your "
+                    "to contain a valid PySide or PyQt install. Either install PySide into your "
```
julien-lang, line 223:
```suggestion
                    "to contain a valid PySide2 or PySide6 install. Either install PySide2/6 into your "
```

python/tk_desktop/desktop_window.py
```diff
@@ -664,13 +664,11 @@ def _handle_close_event(self):
...
             # message, do not show it a second time.
-            # In Qt4, we can't have a custom icon, so don't display
-            # anything. All the stock Qt icons make it look worse.
-            if sgtk.util.is_macos() or QtCore.qVersion()[0] == "4":
+            if sgtk.util.is_macos():
                 icon = self.systray.NoIcon
             else:
                 icon = self._app_icon
-            # Note that on macOS+Qt4, this call will not show anything,
+            # Note that on macOS, this call will not show anything,
```
julien-lang, line 671: Actually, we might want to get rid of the whole comment if proven that it now works on macOS. 
Can you try it to confirm?

python/tk_desktop/extensions/__init__.py
```diff
@@ -23,23 +23,8 @@
...
-
-            use_mocked_osutils = False
-        except Exception as e:
-            logger.warning("Could not load osutils: %s", e, exc_info=True)
-    else:
-        # PySide 2 build of desktop does not require help
-        # making the application go in the foreground/background, so we
-        # don't need to import an extension module.
-        pass
+    except Exception as e:
```
carlos-villavicencio-adsk, line 26: I'd say we can entirely get rid of the try block.

julien-lang (approved):
LGTM!

python/tk_desktop/extensions/__init__.py
```diff
@@ -21,25 +21,7 @@
...
-
-            use_mocked_osutils = False
-        except Exception as e:
-            logger.warning("Could not load osutils: %s", e, exc_info=True)
-    else:
-        # PySide 2 build of desktop does not require help
-        # making the application go in the foreground/background, so we
-        # don't need to import an extension module.
-        pass
+    import PySide2  # noqa
```
carlos-villavicencio-adsk, line 24: If I look at the rest of the file, this import does nothing. Similar to what we have on the `else` block (for Windows). 

I updated `osutils` last time for macOS and I can confirm this is the only OS when we make use of it. 

I suggest to only leave the `if sys.platform == "darwin":` block. Sorry about asking many changes on this.

carlos-villavicencio-adsk (approved):
Thank you for this cleanup!

julien-lang (approved):

PR #175, opened 2024-07-04
SG-35325 Prevent flaky tests

carlos-villavicencio-adsk:
Increase the existing error handling to have more use cases. This will prevent flaky tests on CI.

### DIFF
--- a/python/tk_desktop/rpc.py
+++ b/python/tk_desktop/rpc.py
@@ -17,7 +17,7 @@ import threading
 import time
 import traceback
 import multiprocessing.connection
-from  multiprocessing.context import AuthenticationError
+from multiprocessing.context import AuthenticationError
 
 # We have to import Python's pickle to serialize our data.
 # tk-core's sgtk.util.pickle module assumes too much about the data
--- a/python/tk_desktop/rpc.py
+++ b/python/tk_desktop/rpc.py
@@ -272,7 +272,7 @@ class RPCServerThread(threading.Thread):
                     )
                     ready = True
                 except AttributeError as e:
-                    # The server has been closed.
+                    # The server has been closed. `_listener` is None.
                     logger.debug("Error during select:", exc_info=True)
                     if self.server._listener:
                         raise
@@ -293,7 +293,7 @@ class RPCServerThread(threading.Thread):
                     )
                     ready = len(rd) > 0
                 except AttributeError as e:
-                    # The server has been closed.
+                    # The server has been closed. `_listener` is None.
                     logger.debug("Error during select:", exc_info=True)
                     if self.server._listener:
                         raise
@@ -469,8 +469,8 @@ class RPCProxy(object):
                 if self._closed:
                     raise RuntimeError("client closed while waiting for a response")
 
-                # Check if there's a WinError that causes flaky behavior on CI
-                if sys.platform == "win32":
+                # Check if there's a WinError that causes some flaky behavior on CI
+                if sys.platform == "win32" and "The handle is invalid" in str(e):
                     import _winapi
 
                     if e.winerror:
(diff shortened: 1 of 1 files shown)


python/tk_desktop/rpc.py
```diff
@@ -280,10 +287,24 @@ def run(self):
...
                 # can use select on osx and linux
-                (rd, _, _) = select.select(
-                    [self.server._listener._socket], [], [], self.LISTEN_TIMEOUT
-                )
-                ready = len(rd) > 0
+                try:
+                    (rd, _, _) = select.select(
+                        [self.server._listener._socket], [], [], self.LISTEN_TIMEOUT
+                    )
+                    ready = len(rd) > 0
```
eduardoChaucaGallegos, line 294: Is it necessary the comparison `len(rd) > 0` ? Because I think that it is possible to use just `len(rd)`

python/tk_desktop/rpc.py
```diff
@@ -443,17 +464,17 @@ def call(self, name, *args, **kwargs):
...
             except OSError as e:
+                # IOError it's an alias of OSError
                 if self._closed:
-                    raise RuntimeError(
-                        "client closed while waiting for a response with OSError exception"
-                    )
+                    raise RuntimeError("client closed while waiting for a response")
+
+                # Check if there's a WinError that causes some flaky behavior on CI
+                if sys.platform == "win32" and "The handle is invalid" in str(e):
```
eduardoChaucaGallegos, line 473: I prefer that the string "The handle is invalid" be a variable, for example a class variable, so:
`if sys.platform == self.WIN32 and self._INVALID_ HANDLE_MESSAGE in str(e):`

same for "win32" because it is used plenty times in this class

python/tk_desktop/rpc.py
carlos-villavicencio-adsk: It seems `ready` was defined as boolean type so this is to keep type consistency.

python/tk_desktop/rpc.py
```diff
@@ -264,12 +265,18 @@ def run(self):
...
             # test to see if there is a connection waiting on the pipe
-            if sys.platform == "win32":
+            if is_windows():
                 try:
                     mpc_win32.WaitNamedPipe(
                         self.server.address, self.LISTEN_TIMEOUT * 1000
                     )
                     ready = True
+                except AttributeError as e:
+                    # The server has been closed. `_listener` is None.
```
julien-lang, line 275: Minor point, I think this comment should be moved to the if block right below.

python/tk_desktop/rpc.py
```diff
@@ -280,10 +287,24 @@ def run(self):
...
+                    )
+                    ready = len(rd) > 0
+                except AttributeError as e:
+                    # The server has been closed. `_listener` is None.
+                    logger.debug("Error during select:", exc_info=True)
+                    if self.server._listener:
+                        raise
+                    ready = False
+                except select.error as e:
+                    import errno
```
julien-lang, line 302: I always wonder about those. Should we import everything at the beginning of the file or at "runtime"?

python/tk_desktop/rpc.py
```diff
@@ -349,7 +370,7 @@ def run(self):
                         logger.debug("   traceback:\n%s" % traceback.format_exc())
                         if respond:
                             connection.send(pickle.dumps(e))
-            except (EOFError, IOError):
+            except (EOFError, IOError, AuthenticationError) as e:
```
julien-lang, line 373: unused  variable
```suggestion
            except (EOFError, IOError, AuthenticationError):
```

julien-lang (commented):
minor points. No blockers

python/tk_desktop/rpc.py
carlos-villavicencio-adsk: I think I'm going to stick to PEP 8. We don't have a circular import problem here, and the import is not expensive, so I don't see why not put it at the beginning of the file. Moving it.

julien-lang (approved):
LGTM!

PR #184, opened 2024-12-11
SG-36489 Fix aboutbox/licenses encoding

carlos-villavicencio-adsk:
Provide a stronger Unicode reading

### DIFF
--- a/python/tk_desktop/licenses.py
+++ b/python/tk_desktop/licenses.py
@@ -27,5 +27,5 @@ class Licenses(QtGui.QDialog):
         self.ui = licenses.Ui_Licenses()
         self.ui.setupUi(self)
         # setSource seems broken on Qt4, so use setHtml instead.
-        with open(LICENSE_LOCATION) as f:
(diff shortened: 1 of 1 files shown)


julien-lang (approved):
LGTM!

python/tk_desktop/licenses.py
```diff
@@ -27,5 +27,5 @@ def __init__(self, parent=None):
         self.ui = licenses.Ui_Licenses()
         self.ui.setupUi(self)
         # setSource seems broken on Qt4, so use setHtml instead.
-        with open(LICENSE_LOCATION) as f:
+        with open(LICENSE_LOCATION, encoding="utf-8", errors="replace") as f:
```
julien-lang, line 30: I don't think we should add the error parameter because we distribute this file in tk-desktop so we don't expect encoding errors

```suggestion
        with open(LICENSE_LOCATION, encoding="utf-8") as f:
```

julien-lang (approved):

PR #187, opened 2025-01-27
SG-36162 Refactor __launch_app_proxy_for_project in a QThread

carlos-villavicencio-adsk:
Move the logic for phase 1 (update last accessed timestamp in SG)  when launching of the app proxy for a project in a `QThread` ~~[Background Task Manager](https://developers.shotgridsoftware.com/tk-framework-shotgunutils/task_manager.html)~~ so it doesn't block the UI.

### DIFF
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -1362,7 +1362,9 @@ class DesktopWindow(SystrayWindow):
             self.requested_pipeline_configuration_id = requested_pipeline_configuration_id
 
             self.project_overlay.start_progress()
-            # trigger an update to the model to track this project access
+            # Trigger an update to the model to track this project access
+            # this is done in a separate thread to avoid blocking the UI.
+            # After it finishes, we start phases 2 and 3 in __get_pipeline_configurations method
             self.set_just_accessed_thread = StartAppProxyForProject(self.__set_project_just_accessed, project)
             self.set_just_accessed_thread.finished.connect(self.__get_pipeline_configurations)
             self.set_just_accessed_thread.start()
@@ -1773,6 +1775,9 @@ class DesktopWindow(SystrayWindow):
 
 
 class StartAppProxyForProject(QtCore.QThread):
+    """
+    Thread to set the accessed time for a project.
+    """
     def __init__(self, fn, project):
         super().__init__()
         self.project = project
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -1359,11 +1359,31 @@ class DesktopWindow(SystrayWindow):
             self.ui.actionRefresh_Projects.setVisible(False)
 
             self.current_project = project
+            self.requested_pipeline_configuration_id = requested_pipeline_configuration_id
 
+            self.project_overlay.start_progress()
             # trigger an update to the model to track this project access
-            self.__set_project_just_accessed(project)
-            QtGui.QApplication.instance().processEvents()
+            self.set_just_accessed_thread = StartAppProxyForProject(self.__set_project_just_accessed, project)
+            self.set_just_accessed_thread.finished.connect(self.__get_pipeline_configurations)
+            self.set_just_accessed_thread.start()
+            # self.__set_project_just_accessed(project)
+            # QtGui.QApplication.instance().processEvents()
+        except Exception as error:
+            log.exception(str(error))
+            message = (
+                "%s"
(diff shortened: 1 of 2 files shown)


python/tk_desktop/desktop_window.py
```diff
@@ -1753,3 +1778,18 @@ def handle_about(self):
 
         about = AboutScreen(parent=self, body=body)
         about.exec_()
+
+
+class StartAppProxyForProject(QtCore.QThread):
```
julien-lang, line 1783: This is a quite generic class. Maybe we want a more generic name?

But do we even need a class for running a simple callback?

python/tk_desktop/desktop_window.py
```diff
@@ -1359,11 +1359,37 @@ def __launch_app_proxy_for_project(
...
+            log.exception(str(error))
+            message = (
+                "%s"
+                "\n\nThere was an error connecting to Flow Production Tracking."
+                "\n\nFor more details, see the console." % str(error)
+            )
+            self.project_overlay.show_error_message(message)
+            return
 
+    def __get_pipeline_configurations(self):
```
julien-lang, line 1387: I'm not sure I like the name of this method. Any reason for it?

Why not something like `__launch_app_proxy_for_project_follow_up`?

python/tk_desktop/desktop_window.py
barbara-darkshot: hmm instead of creating a new class, would it make sense to use the `BackgroundTaskManager`?
you have an instance created [here](https://github.com/shotgunsoftware/tk-desktop/blob/master/python/tk_desktop/desktop_engine_site_implementation.py#L58-L59) and your DesktopWindow object is created in the [same file](https://github.com/shotgunsoftware/tk-desktop/blob/master/python/tk_desktop/desktop_engine_site_implementation.py#L510) so you should be able to easily pass the BGTaskManager object as parameter to the class constructor and use its signals to do all the different actions
I haven't tested anything so I'm not sure it could work, just an idea :)

carlos-villavicencio-adsk: @barbara-darkshot sounds like a good idea. I'll try out!

carlos-villavicencio-adsk: @barbara-darkshot I think it was possible to use the task manager here. When testing everything looks similar to using QThreads. Can you take a look and check if anything look out of place?

python/tk_desktop/desktop_window.py
```diff
@@ -1753,3 +1780,23 @@ def handle_about(self):
...
+        self.__launch_app_proxy_for_project_follow_up()
+
+    def _on_task_failed(self, task_id, search_id, msg, stack_trace):
+        """
+        Callback when a task has failed by the BackgroundTaskManager.
+        """
+        if task_id != self.__set_project_just_accessed_task:
+            return
+
+        log.error(f"Launching app proxy for project failed. Message: {msg}")
```
barbara-darkshot, line 1802: you also need to clear `self.__set_project_just_accessed_task` if the task has failed
maybe you can also change the error message to have a better reflection of what has failed by replacing `Launching app proxy for project failed` by `Error happened when loading pipeline configuration: {msg}`
(tbh, as a user/artist, I don't know what is the app proxy :D )
also wondering if we could use `self.project_overlay.show_error_message(message)` to display the error message to the user as well?

barbara-darkshot (changes requested):
looking good!
just requesting minor changes but happy to discuss

python/tk_desktop/desktop_window.py
```diff
@@ -218,6 +217,11 @@ def __init__(self, console, parent=None):
         # Setup the console
         self.__console = console
 
+        # Task Manager
+        self._bg_task_manager = task_manager
+        self._bg_task_manager.task_completed.connect(self._on_task_completed)
+        self._bg_task_manager.task_failed.connect(self._on_task_failed)
```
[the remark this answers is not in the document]
carlos-villavicencio-adsk: Good call

python/tk_desktop/desktop_window.py
carlos-villavicencio-adsk: I've refactored the error handling on the callback and cleared the task. Good catch!

barbara-darkshot (approved):
LGTM!

julien-lang (approved):

julien-lang (approved):

eduardoChaucaGallegos (approved):

PR #193, opened 2025-03-14
SG-38672 Removed dead authentication code

eduardoChaucaGallegos:
### **Description**
This PR removes dead code related to uses_legacy_authentication.

This cleanup follows a previously identified issue where LooseVersion("v1.9.1") was parsed as ["v1", 9, 1], which is described in more detail in that earlier [PR](https://github.com/shotgunsoftware/tk-desktop/pull/194).

Rather than removing the unused code at the time, we opted to fix the issue by updating the version string to avoid the problematic comparison. This PR now completes the process by removing the obsolete logic that was originally tied to that condition.

### **Test**
image

### DIFF
--- a/python/tk_desktop/desktop_engine_site_implementation.py
+++ b/python/tk_desktop/desktop_engine_site_implementation.py
@@ -442,9 +442,6 @@ class DesktopEngineSiteImplementation(object):
             # DO NOT raise exception. It's reasonnable to log an error about it but
             # we don't want to break normal execution for metric related logging.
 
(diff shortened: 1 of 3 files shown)


carlos-villavicencio-adsk (approved):
LGTM. Thanks for the dead code removal.

carlos-villavicencio-adsk (approved):
I think we need to update the Jira ticket ID on the PR title

julien-lang (approved):

PR #194, opened 2025-03-14
SG-38064 Inconsistent comparison - bug fix

eduardoChaucaGallegos:
### **Description**
- The error happens because LooseVersion("v1.9.1") is parsed as ["v1", 9, 1], while LooseVersion("1.1.0") is ["1", 1, 0]. The comparison fails when Python tries to compare "v1" (a string) with 1 (an integer). Adding the "v" prefix would make them comparable.
image

### **Test**
- With the fix, Desktop launches without any issues.
image


carlos-villavicencio-adsk (approved):
LGTM

PR #195, opened 2025-05-05
SG-38078 hook PreInitialization to ensure early early initialization tasks during bootstrap

eduardoChaucaGallegos:
### **Description**

We identified that importing from core indirectly brings in Qt (specifically PySide6), which causes issues in our environment.
While we explored workarounds such as monkey patching core before import, these solutions proved fragile and highly dependent on specific core versions.
As a result, and despite the downside of code duplication, we've decided to duplicate the necessary logic to avoid importing core directly.
This ensures greater stability and compatibility until a more robust solution becomes viable.

This PR introduces the `hook_pre_initialization` mechanism to handle early initialization tasks during the bootstrap process. The hook ensures that critical tasks, such as importing libraries like `opentimelineio` or others, are performed before PySide6 or other dependencies are loaded, avoiding potential conflicts caused by Qt initialization issues or version mismatches.

**1.   `execute_pre_initialization_hook` Function**

- Dynamically attempts to load and execute the PreInitialization custom hook from a path specified in the **tk-desktop.yml** configuration file, under the **settings.tk-desktop.project.hooks** section.
-  Falls back to the default hook located in the `hooks` directory if the variable is not set.

**2. Default Hook Implementation:**

- Added a default `PreInitialization` hook in the `hooks` directory.
- Includes a clear docstring explaining its purpose, usage, and how to customize it via the environment variable.

**3. Bootstrap Integration:**

- The `execute_pre_initialization_hook` function is called early in the bootstrap process to ensure a stable environment before PySide6 or other dependencies are loaded.

#### Additional NOTE
- This implementation avoids inheriting from `Hook` in `sgtk` to prevent unintended imports of PySide6. Using `sgtk.Hook` would automatically load parts of the `sgtk` framework, which could trigger the initialization of PySide6 or other dependencies too early in the bootstrap process. This early initialization could lead to conflicts or version mismatches, especially with libraries like `opentimelineio` or other Qt-based dependencies.

- By implementing the `PreInitialization` hook as a standalone class without relying on `sgtk.Hook`, we ensure that no unnecessary imports occur, maintaining a clean and conflict-free environment during the early stages of the bootstrap process. This approach provides greater control over the initialization order and avoids hard-to-debug issues caused by improper library setup.

### **Tests**
- Version with problem
image

-A custom hook has been defined in the configuration under the pre_initialization hook setting, pointing to the location of the new custom hook
image

- Executing an example with opentimelineio
image


julien-lang:
@eduardoChaucaGallegos are you sure importing `sgtk.Hook` will trigger the import of Qt/PySide?

python/utils/bootstrap_utilities.py
```diff
@@ -391,3 +392,37 @@ def handle_error(data, proxy=None):
...
+        # Dynamically load the module
+        spec = importlib.util.spec_from_file_location("import_libraries", hook_path)
+        hook_module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(hook_module)
+
+        # Execute the `execute` method of the `ImportLibraries` class
+        hook_class = hook_module.ImportLibraries
+        hook_class.execute()
+    else:
+        from hooks.import_libraries import ImportLibraries
```
julien-lang, line 426: Could you make the default hook imported and executed the same way as the "set" version?

python/utils/bootstrap_utilities.py
```diff
@@ -391,3 +392,37 @@ def handle_error(data, proxy=None):
...
+
+    The purpose of this hook is to import necessary libraries before PySide6 or other
+    dependencies are loaded, ensuring a stable environment and avoiding potential
+    conflicts caused by Qt initialization issues or version mismatches.
+
+    Any exceptions during execution are logged for debugging purposes.
+    """
+
+    # Retrieve the hook path from the environment variable
+    hook_path = os.environ.get("SGTK_HOOK_IMPORT_LIBRARIES")
```
julien-lang, line 414: Can we have a setting in the info.yml too?

hooks/import_libraries.py
```diff
@@ -0,0 +1,42 @@
...
+
+    Example:
+        # Import libraries that need to be loaded early
+        import opentimelineio
+        import some_other_library
+
+    Note:
+        This hook is critical for avoiding hard-to-debug errors caused by improper
+        library initialization.
+    """
```
julien-lang, line 28: I'm not thrilled about the name of this hook. The "import library" part is just an example of what can be done here. But overriding it will make way more.

So I think we should rename it in a more generic way.

python/utils/bootstrap_utilities.py
```diff
@@ -391,3 +392,39 @@ def handle_error(data, proxy=None):
...
+
+    # Retrieve the hook path from the environment variable
+    hook_path = os.environ.get("SGTK_HOOK_PRE_INITIALIZATION")
+
+    # Check if the hook file exists, so if the environment
+    # variable is not set, use the default hook path
+    if not hook_path or not os.path.exists(hook_path):
+        hook_path = os.path.join(
+            os.path.dirname(__file__), "..", "..", "hooks", "pre_initialization.py"
+        )
```
julien-lang, line 421: I think we need a log if the setting is set but not a valid path
```suggestion
    if hook_path and not os.path.exists(hook_path):
        logger.error("No such file %s; reverting to hook's default value")
        hook_path = None

     if not hook_path:
        hook_path = os.path.join(
            os.path.dirname(__file__), "..", "..", "hooks", "pre_initialization.py"
        )
```

We probably need a log line a bit better than my suggestion

python/utils/bootstrap_utilities.py
```diff
@@ -391,3 +392,39 @@ def handle_error(data, proxy=None):
...
+
+    # Dynamically load the module
+    spec = importlib.util.spec_from_file_location("pre_initialization", hook_path)
+    hook_module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(hook_module)
+
+    # Execute the `execute` method of the `PreInitialization` class
+    hook_class = getattr(hook_module, "PreInitialization", None)
+    hook_class.execute()
```
julien-lang, line 430: I would put those lines in a try/except block with a list of exceptions related to not finding the modules/classes/methods/...
Because we don't want to crash the main process if the hook fails.

python/utils/bootstrap_utilities.py
julien-lang: @eduardoChaucaGallegos what about this point?

python/utils/bootstrap_utilities.py
```diff
@@ -391,3 +392,86 @@ def handle_error(data, proxy=None):
...
+    it falls back to dynamically loading the default hook located in the 'hooks'
+    directory of the project.
+
+    The purpose of this hook is to perform early initialization tasks, such as importing
+    necessary libraries, before PySide6 or other dependencies are loaded. This ensures
+    a stable environment and avoids potential conflicts caused by Qt initialization issues
+    or version mismatches.
+    """
+
+    hook_path = os.environ.get("SGTK_HOOK_PRE_INITIALIZATION")
```
eduardoChaucaGallegos, line 438: @julien-lang, did you mean something like this on info.yml ?
```
hooks:
  pre_initialization: /path/to/custom/pre_initialization.py
```

julien-lang: Yes, something like that.

info.yml
```diff
@@ -137,7 +137,13 @@ configuration:
...
                      engine."
 
-
+    hook_pre_initialization:
+        type: hook
+        default_value: pre_initialization
+        description: "This hook is called before the engine is initialized. It is
+                     passed the engine instance and the environment configuration
+                     object. This hook can be used to modify the environment
+                     configuration before it is used to initialize the engine."
```
julien-lang, line 146: I'm 90% sure we can get rid of the quotes here

```suggestion
        description: This hook is called before the engine is initialized. It is
                     passed the engine instance and the environment configuration
                     object. This hook can be used to modify the environment
                     configuration before it is used to initialize the engine.
```

python/utils/bootstrap_utilities.py
```diff
@@ -391,3 +392,160 @@ def handle_error(data, proxy=None):
...
+    sys.path.insert(0, data["core_python_path"])
+    import sgtk
+
+    # Set up logging with the rpc.
+    proxy = _create_proxy(data)
+    handler = ProxyLoggingHandler(proxy)
+    sgtk.LogManager().initialize_custom_handler(handler)
+    logger = sgtk.LogManager.get_logger(__file__)
+    logger.error(message)
+    proxy.close()
```
julien-lang, line 419: Not super important, but I find this function awfully wasteful when called multiple times. 
Is there a way to keep things open for the duration of ... the process. Like with a decorator function maybe or something similar?

python/utils/bootstrap_utilities.py
```diff
@@ -391,3 +392,160 @@ def handle_error(data, proxy=None):
...
+        # Step 3: Fallback to the default hook if no custom hook is defined
+        if hook_path and not os.path.exists(hook_path):
+            message = (
+                f"Custom pre-initialization hook specified is set to '{hook_path}',"
+                " but the file does not exist. "
+                "Falling back to the default pre-initialization hook. pipiree"
+            )
+            logger_via_proxy(data, message)
+            hook_path = None
+
```
julien-lang, line 525: Looks good but I still don't understand why we can't call methods of tk-core code here instead of duplicating the code here?

I know that's because PySide6 (QtImporter) is called early on but I reviewed the code about Qt imports recently and there is definitely a big part of core that is supposed to be used **without** Qt/PySide

python/utils/bootstrap_utilities.py
```diff
@@ -391,3 +393,163 @@ def handle_error(data, proxy=None):
...
+    proxy.close()
+
+
+def parse_yml_file(file_path):
+    """
+    Parses a simple YAML file into a dictionary without using the `yaml` library.
+
+    This function was created to avoid importing the `yaml` library due to known
+    compatibility issues with PySide6. It supports basic nested structures and is
+    designed to work specifically for the needs of this hook.
```
carlos-villavicencio-adsk, line 431: Which yaml library is having issues with PySide6? PyYaml or Ruamel? Can you elaborate more on these compatibility issues? I don't like the idea of having an implementation for parsing a yaml file on our own.

tests/test_about_box.py
```diff
@@ -63,16 +64,28 @@ def test_3rd_party_links(licence_file_links):
...
+                    url,
+                    headers={"Accept-Language": "en", "User-Agent": "Mozilla/5.0"},
+                )
 
-            contents = request.urlopen(r).read()
-        except Exception as e:
-            raise pytest.fail("Failed to open {0}, error: {1}".format(url, e))
+                contents = request.urlopen(r).read()
+                break
+            except Exception as e:
```
carlos-villavicencio-adsk, line 80: [nitpick] if possible please use a more specific error class.

carlos-villavicencio-adsk (commented):
Added a couple observations/questions.

python/utils/bootstrap_utilities.py
eduardoChaucaGallegos: Good catch! After a quick test, importing `yaml ` from `tank_vendor `doesn’t import PySide6, so it should be safe to use it here.

python/utils/bootstrap_utilities.py
```diff
@@ -391,3 +392,128 @@ def handle_error(data, proxy=None):
...
+            info = yaml.load(fh, Loader=yaml.FullLoader)
+        default_hook_name = (
+            info.get("configuration", {})
+            .get("hook_pre_initialization", {})
+            .get("default_value", "")
+        )
+
+        # Step 2: Read tk-desktop.yml to check for a custom hook
+        tk_desktop_yml_path = os.path.join(
+            data.get("config_path"),
```
julien-lang, line 467: @eduardoChaucaGallegos using this `config_path`, you can easily implement the missing `{config}` pattern in this method, isn't it?

julien-lang (approved):

carlos-villavicencio-adsk (approved):

PR #201, opened 2025-07-16
SG-39698 Removed python 2 try/except sgutils block

eduardoChaucaGallegos:


python/tk_desktop/rpc.py
```diff
@@ -214,8 +209,13 @@ def __init__(self, engine, authkey=None):
             family = "AF_PIPE"
         else:
             family = "AF_UNIX"
+        self.authkey = (
+            self.authkey.encode("utf-8")
+            if isinstance(self.authkey, str)
+            else self.authkey
+        )
```
julien-lang, line 216:
```suggestion
        if isinstance(self.authkey, str)
            self.authkey = self.authkey.encode("utf-8")
```

julien-lang (approved):

PR #211, opened 2026-02-03
SG-39788 Fix data persistence issue after re-authentication

carlos-villavicencio-adsk:
## Problem
When a user re-authenticates with a different account after session expiry, the Flow Production Tracking Desktop UI was still showing the previous user's name, email, and projects - a security concern.

## Security Impact
- Information Disclosure: New user can see project names they shouldn't have access to
- User Confusion: UI shows wrong user's name/email
- Data Integrity: User-specific fields (favorites, last accessed) show incorrect data

## How the Fix Works
1. Reactive Detection: When refresh_user_credentials() is called (before launching apps, opening folders, etc.), the fix detects if a different user authenticated and updates the UI immediately.
2. Proactive Validation: When the project model data is refreshed or when the user navigates back to the projects view, the fix validates that the cached user matches the current authenticated user and refreshes the UI if there's a mismatch.
3. Complete UI Refresh: When a user change is detected:
    - User name and site URL in the menu are updated
    - User thumbnail/avatar is updated
    - Project model is hard-refreshed to show correct projects for the new user


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -529,11 +529,77 @@ def get_current_user(self):
...
+                # First time check or no previous user - just update the cached info
+                # to ensure it's in sync with the authenticated user
+                if self._user is None or self._user.login != current_login:
+                    self._user = human_user
+                    self._current_login = self._engine.sgtk.shotgun.find_one(
+                        "HumanUser", [["login", "is", current_login]], ["id", "login"]
+                    )
+
+        except Exception:
+            logger.exception("Error checking for user change after credential refresh.")
```
julien-lang: I'm not crazy about surouding entire functions by global try/catch. Can we improve that?

python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -529,11 +529,77 @@ def get_current_user(self):
...
+
+                # Update cached user information
+                self._user = human_user
+                self._current_login = self._engine.sgtk.shotgun.find_one(
+                    "HumanUser", [["login", "is", current_login]], ["id", "login"]
+                )
+
+                # Notify the desktop window to refresh user info and projects
+                if hasattr(self, "desktop_window") and self.desktop_window is not None:
+                    self.desktop_window.on_user_changed()
```
julien-lang, line 586: Ok so that's the fix. I mean calling this method in the right context fixes the issue.

But we are in the "site view" here. What about the project view? Do we have to fix it there too?

carlos-villavicencio-adsk: In that case, it's already handled. 
- If the project can be accessed by the incoming user, then it's all good.
- If the project cannot be accessed by the incoming user, then an error is shown, and the user can go back to the site view to select a project they can access.

Image

julien-lang (approved):

eduardoChaucaGallegos (approved):

PR #212, opened 2026-03-10
SG-42559 Load osutils from app bundle via sys.prefix

julien-lang:
## Summary

Since SGD 3.0.0 the compiled `osutils` macOS C extension lives inside the FPT
Desktop app bundle rather than being committed as a binary blob to this public
repository.

Companion change in `tk-desktop-internal`: the build script places the compiled
`.so` under `{bundle}/Contents/Resources/Desktop/extensions/darwin_python3/` and
explicitly sets `ARCHFLAGS="-arch x86_64 -arch arm64"` for a universal binary.

## Changes

**`python/tk_desktop/extensions/__init__.py`**

- Extracted loading logic into `_load_osutils_from_bundle()` helper function.
- The helper derives the bundle path from `sys.prefix` (always
  `{App}/Contents/Resources/Python3` inside the Desktop app) - no environment
  variable required.
- Falls back to the pre-compiled binaries committed to `darwin_python3/`
  (Python 3.7-3.11) for backward compatibility with older installers.

## Loading strategy

1. Try bundle: `{sys.prefix}/../Desktop/extensions/darwin_python3/osutils.cpython-XY-darwin.so`
2. Try legacy: `from .darwin_python3 import osutils` (Python 3.7-3.11 binaries in this repo)
3. Fall back to no-op mock class (non-macOS, or missing bundle)

## Testing

- Outside Desktop app (developer venv): `sys.prefix` points elsewhere,
  `os.path.isdir()` returns `False`, falls through to legacy or mock. No regression.
- Inside Desktop app 3.0.0+: bundle path resolves, correct ABI tag matched.

Jira: [SG-42559](https://jira.autodesk.com/browse/SG-42559)


carlos-villavicencio-adsk (approved):
LGTM. However, there are some copilot feedback that can be useful to review.

PR #213, opened 2026-03-24
SG-43220 Fix Desktop About to show the selected project is config

veinfx:
## Summary

Display the selected project configuration in Desktop About.

## Problem

Desktop About incorrectly displays the site configuration instead of the selected project configuration.

## Change

Use the selected project config first.
If it is not available, use the current engine config.

## Result

Desktop About shows the selected project config more clearly.

project_config_about


carlos-villavicencio-adsk (approved):
LGTM! Thanks for the contribution

PR #218, opened 2026-05-07
SG-42329 Fix multiple apps in desktop list after restart

carlos-villavicencio-adsk:
## Problem
Every time "Reload and Restart" was clicked in the Flow Production Tracking desktop app, the app launcher panel duplicated all of its buttons

After this | We end up like this
--|--
Screenshot 2026-05-07 115726 | Screenshot 2026-05-07 120014

## Root Cause
The desktop app communicates with the project subprocess over RPC. There are two ways a connection can end, and they follow different code paths — only one of which clears the launcher UI.

## The Fix
One line changed in _on_proxy_created. Instead of only clearing context-menu actions, it now calls clear_app_uis() — the same full reset that _on_proxy_closing performs — ensuring the launcher is always empty before the incoming engine registers its commands.

**Does it make sense to have both `clear_actions_from_project_menu` and `clear_app_uis` invoked? Or is `clear_app_uis` enough?**

- clear_app_uis() alone is enough — calling clear_actions_from_project_menu() afterwards would be a no-op. 
- reset() destroys and recreates the QMenu entirely. After that, calling clear_actions_from_project_menu() → clear_actions() would iterate over an already-empty menu and do nothing.

### DIFF
--- a/tests/test_engine_user_refresh.py
+++ b/tests/test_engine_user_refresh.py
@@ -557,17 +557,3 @@ class TestProxyLifecycle:
         DesktopEngineSiteImplementation._on_proxy_closing(engine_impl)
 
         engine_impl.desktop_window.clear_app_uis.assert_called_once()
-
-    def test_subprocess_initiated_reload_clears_ui_via_proxy_created(self):
-        """
-        When reload is initiated by the subprocess (destroy_app_proxy path),
-        proxy_closing is never emitted — only proxy_created fires.  Verify that
-        clear_app_uis is still called in that scenario, covering the original
-        duplication bug.
-        """
-        engine_impl = self._make_engine_impl()
-
-        # proxy_closing does NOT fire on the subprocess-initiated reload path.
-        DesktopEngineSiteImplementation._on_proxy_created(engine_impl)
-
-        engine_impl.desktop_window.clear_app_uis.assert_called_once()
--- a/tests/test_engine_user_refresh.py
+++ b/tests/test_engine_user_refresh.py
@@ -467,107 +467,107 @@ class TestEngineUserRefreshIntegration:
         engine_impl.desktop_window.on_user_changed.assert_not_called()
 
 
-# class TestProxyLifecycle:
-#     """
-#     Tests for the proxy connection lifecycle handlers.
+class TestProxyLifecycle:
+    """
+    Tests for the proxy connection lifecycle handlers.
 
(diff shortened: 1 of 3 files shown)


tests/test_engine_user_refresh.py
```diff
@@ -465,3 +465,109 @@ def test_same_user_reauth_no_window_notification(self, mock_sgtk, mock_auth):
...
+        engine_impl = self._make_engine_impl()
+
+        DesktopEngineSiteImplementation._on_proxy_closing(engine_impl)
+
+        engine_impl.desktop_window.clear_app_uis.assert_called_once()
+
+    def test_subprocess_initiated_reload_clears_ui_via_proxy_created(self):
+        """
+        When reload is initiated by the subprocess (destroy_app_proxy path),
+        proxy_closing is never emitted — only proxy_created fires.  Verify that
```
chenm1adsk:
```suggestion
        proxy_closing is never emitted — only proxy_created fires. Verify that
```
There is a extra space in this line.

tests/test_engine_user_refresh.py
```diff
@@ -465,3 +465,109 @@ def test_same_user_reauth_no_window_notification(self, mock_sgtk, mock_auth):
...
+        _on_proxy_closing must still call clear_app_uis() — existing behaviour
+        must not have been broken by the fix.
+        """
+        engine_impl = self._make_engine_impl()
+
+        DesktopEngineSiteImplementation._on_proxy_closing(engine_impl)
+
+        engine_impl.desktop_window.clear_app_uis.assert_called_once()
+
+    def test_subprocess_initiated_reload_clears_ui_via_proxy_created(self):
```
chenm1adsk: Hey Charlie, I have a question - I am wondering how does this function mimic "When reload is initiated by the subprocess (destroy_app_proxy path)"? It looks exactly the same as `test_on_proxy_created_clears_full_app_ui`

chenm1adsk (commented):
Left a small suggested change and a question. The remaining looks good to me!

yungsiow (approved):
LGTM!

tests/test_engine_user_refresh.py
carlos-villavicencio-adsk: Good catch. I've removed the repeated test.

chenm1adsk (approved):
LGTM!

PR #219, opened 2026-05-11
SG-41322 Fix greyed-out title bar close button on Windows

stevelittlefish:
## Problem

On Windows, the native title bar close (X) button is greyed out from the moment the Toolkit desktop app launches, so the window cannot be closed via the X.

## Root cause

`DesktopWindow.set_on_top` is called at startup from `_load_settings` (with `value=False`) before the window has been shown for the first time. It used the plural `setWindowFlags(flags & ~Qt.WindowStaysOnTopHint)`, which **replaces** the entire flag set rather than toggling a single bit.

Before the window is realized, `self.windowFlags()` on Windows does not yet include `Qt.WindowCloseButtonHint` in the returned bitmask — that hint is part of the implicit `Qt.Window` default that Qt resolves at window realization. Re-applying the incomplete bitmask via the plural setter caused Qt to render the title bar without an enabled close button.

The same code path runs whenever the user toggles "Keep on Top", so the bug was also reproducible by toggling that menu item.

## Fix

Switched to the singular `setWindowFlag(Qt.WindowStaysOnTopHint, value)` (Qt 5.9+), which toggles only that one bit and preserves every other implicit hint, including `WindowCloseButtonHint`.

## Testing

- Launch the app fresh on Windows — title bar X is enabled and closes/hides the window as expected.
- Toggle "Keep on Top" on, then off — X remains enabled in both states.
- macOS / Linux unaffected (no change in observable behavior).

Jira: [SG-41322](https://jira.autodesk.com/browse/SG-41322)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -13,7 +13,9 @@
 #
 [run]
 source=.
-omit=tests/*
+omit=
+    tests/*
+    python/tk_desktop/desktop_window.py
 
 [report]
 exclude_lines =
--- a/python/tk_desktop/desktop_window.py
+++ b/python/tk_desktop/desktop_window.py
@@ -1179,17 +1179,11 @@ class DesktopWindow(SystrayWindow):
         return self.windowFlags() & QtCore.Qt.WindowStaysOnTopHint
 
     def set_on_top(self, value):
-        flags = self.windowFlags()
         visible = self.isVisible()
 
-        if value:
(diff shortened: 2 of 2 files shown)


carlos-villavicencio-adsk (approved):
LGTM

yungsiow (approved):
LGTM!

carlos-villavicencio-adsk (approved):
LGTM

PR #221, opened 2026-06-10
SG-43676 Handle popen and stop processes on application shutdown

joel-fortin:
**Problem**
The bootstrap python.exe process spawned when opening a project was never terminated when Desktop quit, leaving orphan processes that locked Toolkit cache files, preventing tk-core to refresh the cache when needed.

**Change**

- Track the bootstrap Popen handle when Desktop opens a project (launch_python hook) instead of fire-and-forget spawning.
- Terminate the bootstrap subprocess on existing shutdown paths (site_comm.shut_down()): Quit Desktop, back to projects, switch project, and engine destroy.
- Add terminate_process() helper with graceful terminate() → kill() fallback, plus unit tests for the termination logic.
- Fixes orphan Shotgun\Python3\python.exe processes that could survive after Desktop quit and lock Toolkit cache files (e.g. pkgs.zip), blocking manual cache cleanup and config refresh.

Note: Closing the dialog to the system tray is unchanged — the subprocess remains running while a project is active in the background.

Testing Done:

- Open Desktop → open a project → confirm apps load normally
- Confirm bootstrap python.exe is running (bootstrap.py in command
- Quit Desktop fully → confirm bootstrap process is gone
- Repeat open/quit 3–5 times → confirm no accumulating orphan processes
- Open project → close dialog to tray → confirm bootstrap still running → reopen from tray → project still works
- Back to projects → confirm bootstrap process is gone
- After quit, confirm cache folder is not locked by orphan processes
- Make sure only one python.exe shotgun process is active at all time.

### DIFF
--- a/hooks/launch_python.py
+++ b/hooks/launch_python.py
@@ -59,7 +59,12 @@ class LaunchPython(Hook):
         # will be shared with the child process and it prevent restarting the server
         # after the process closes.
         # Solution was found here: http://stackoverflow.com/a/13593715
-        subprocess.Popen(args, startupinfo=startupinfo, close_fds=True)
+        process = subprocess.Popen(args, startupinfo=startupinfo, close_fds=True)
+
+        # Keep track of the bootstrap process so it can be terminated when the
+        # desktop app disconnects from the project or quits.
+        if hasattr(self.parent, "site_comm"):
+            self.parent.site_comm.set_bootstrap_process(process)
 
     def path_to_bootstrap(self):
         """
--- a/hooks/launch_python.py
+++ b/hooks/launch_python.py
@@ -17,10 +17,9 @@ import sys
 import subprocess
 
 import tank
-from tank import Hook
 
 
-class LaunchPython(Hook):
+class LaunchPython(tank.Hook):
     def execute(self, project_python, pickle_data_path, utilities_module_path):
         """
         Launch the python process that will start the project specific tk-desktop
--- a/hooks/launch_python.py
+++ b/hooks/launch_python.py
@@ -59,7 +59,9 @@ class LaunchPython(Hook):
         # will be shared with the child process and it prevent restarting the server
         # after the process closes.
         # Solution was found here: http://stackoverflow.com/a/13593715
-        process = subprocess.Popen(args, startupinfo=startupinfo, close_fds=True)
+        process = subprocess.Popen(
+            args, startupinfo=startupinfo, close_fds=True
+        )  # nosec B603 - args are built from trusted toolkit internals, not user input
 
         # Keep track of the bootstrap process so it can be terminated when the
         # desktop app disconnects from the project or quits.
--- a/hooks/launch_python.py
+++ b/hooks/launch_python.py
@@ -63,8 +63,10 @@ class LaunchPython(Hook):
 
         # Keep track of the bootstrap process so it can be terminated when the
         # desktop app disconnects from the project or quits.
-        if hasattr(self.parent, "site_comm"):
+        try:
             self.parent.site_comm.set_bootstrap_process(process)
+        except AttributeError:
+            pass
 
     def path_to_bootstrap(self):
         """
--- a/python/tk_desktop/bootstrap_process.py
+++ b/python/tk_desktop/bootstrap_process.py
@@ -1,0 +1,54 @@
+# Copyright (c) 2026 Shotgun Software Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+Helpers for managing the project bootstrap subprocess launched by tk-desktop.
+"""
+
+import subprocess
+
+from sgtk import LogManager
+
+logger = LogManager.get_logger(__name__)
+
+DEFAULT_TERMINATION_TIMEOUT = 5
+
+
+def terminate_process(process, timeout=DEFAULT_TERMINATION_TIMEOUT):
+    """
+    Terminates a subprocess if it is still running.
+
+    :param process: A :class:`subprocess.Popen` instance, or None.
+    :param timeout: Number of seconds to wait for graceful termination before killing.
+    """
+    if process is None:
+        return
+
+    if process.poll() is not None:
+        return
+
+    logger.debug("Terminating bootstrap process (pid %s)...", process.pid)
+    try:
+        process.terminate()
+    except Exception as exc:
+        logger.warning("Error terminating bootstrap process: %s", exc)
+        return
+
+    try:
+        process.wait(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        logger.warning(
+            "Bootstrap process (pid %s) did not exit, killing...", process.pid
+        )
+        try:
+            process.kill()
+            process.wait(timeout=timeout)
+        except Exception as exc:
+            logger.warning("Error killing bootstrap process: %s", exc)
--- a/python/tk_desktop/bootstrap_process.py
+++ b/python/tk_desktop/bootstrap_process.py
@@ -12,7 +12,7 @@
 Helpers for managing the project bootstrap subprocess launched by tk-desktop.
 """
 
-import subprocess
+import subprocess  # nosec B404
 from typing import Optional
 
 from sgtk import LogManager
--- a/python/tk_desktop/bootstrap_process.py
+++ b/python/tk_desktop/bootstrap_process.py
@@ -13,17 +13,17 @@ Helpers for managing the project bootstrap subprocess launched by tk-desktop.
 """
 
 import subprocess  # nosec B404
-from typing import Optional
+import typing
 
-from sgtk import LogManager
+import sgtk
 
-logger = LogManager.get_logger(__name__)
+logger = sgtk.LogManager.get_logger(__name__)
 
 DEFAULT_TERMINATION_TIMEOUT: int = 5
 
 
 def terminate_process(
-    process: Optional[subprocess.Popen],
+    process: typing.Optional[subprocess.Popen],
     timeout: int = DEFAULT_TERMINATION_TIMEOUT,
 ) -> None:
     """
--- a/python/tk_desktop/bootstrap_process.py
+++ b/python/tk_desktop/bootstrap_process.py
@@ -13,15 +13,19 @@ Helpers for managing the project bootstrap subprocess launched by tk-desktop.
 """
 
 import subprocess
+from typing import Optional
 
 from sgtk import LogManager
 
 logger = LogManager.get_logger(__name__)
 
-DEFAULT_TERMINATION_TIMEOUT = 5
+DEFAULT_TERMINATION_TIMEOUT: int = 5
 
 
-def terminate_process(process, timeout=DEFAULT_TERMINATION_TIMEOUT):
+def terminate_process(
+    process: Optional[subprocess.Popen],
+    timeout: int = DEFAULT_TERMINATION_TIMEOUT,
+) -> None:
     """
     Terminates a subprocess if it is still running.
 
--- a/python/tk_desktop/bootstrap_process.py
+++ b/python/tk_desktop/bootstrap_process.py
@@ -41,18 +41,26 @@ def terminate_process(
     logger.debug("Terminating bootstrap process (pid %s)...", process.pid)
     try:
         process.terminate()
-    except Exception as exc:
+    except OSError as exc:
         logger.warning("Error terminating bootstrap process: %s", exc)
+
+    try:
+        process.wait(timeout=timeout)
         return
+    except subprocess.TimeoutExpired:
+        logger.warning(
+            "Bootstrap process (pid %s) did not exit after terminate(), killing...",
+            process.pid,
+        )
+
+    try:
+        process.kill()
+    except OSError as exc:
+        logger.warning("Error killing bootstrap process: %s", exc)
 
     try:
         process.wait(timeout=timeout)
     except subprocess.TimeoutExpired:
         logger.warning(
-            "Bootstrap process (pid %s) did not exit, killing...", process.pid
+            "Bootstrap process (pid %s) did not exit after kill().", process.pid
         )
-        try:
-            process.kill()
-            process.wait(timeout=timeout)
-        except Exception as exc:
-            logger.warning("Error killing bootstrap process: %s", exc)
--- a/python/tk_desktop/site_communication.py
+++ b/python/tk_desktop/site_communication.py
@@ -12,6 +12,8 @@
 Implements communication channels between the desktop app and the background process.
 """
 
+import subprocess
+
 from sgtk.platform.qt import QtCore
 from .bootstrap_process import terminate_process
 from .communication_base import CommunicationBase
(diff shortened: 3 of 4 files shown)


tests/test_bootstrap_process.py
```diff
@@ -0,0 +1,60 @@
...
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import subprocess
+import sys
+import os
```
julien-lang: alphabetical order please for imports.

tests/test_bootstrap_process.py
```diff
@@ -0,0 +1,60 @@
...
+from unittest.mock import Mock, MagicMock, patch
```
julien-lang: Can you simply import `unittest.mock` and then access the classes/modules with
```
unittest.mock.Mock...
```

tests/test_bootstrap_process.py
```diff
@@ -0,0 +1,60 @@
...
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "python"))
+
+with patch.dict(
+    "sys.modules",
+    {"sgtk": MagicMock(LogManager=MagicMock(get_logger=MagicMock(return_value=Mock())))},
+):
+    from tk_desktop.bootstrap_process import terminate_process
```
julien-lang: Ohh this is interesting!

We never clean up the sys.path. Maybe we should!

python/tk_desktop/site_communication.py
```diff
@@ -31,6 +32,31 @@ class SiteCommunication(QtCore.QObject, CommunicationBase):
...
+        process = self._bootstrap_process
+        self._bootstrap_process = None
+        terminate_process(process)
+
+    def shut_down(self):
+        """
+        Disconnects from the project proxy and terminates the bootstrap subprocess.
+        """
+        CommunicationBase.shut_down(self)
+        self._terminate_bootstrap_process()
```
julien-lang: When is this call?

python/tk_desktop/site_communication.py
```diff
@@ -31,6 +32,31 @@ class SiteCommunication(QtCore.QObject, CommunicationBase):
...
+        self._bootstrap_process = process
+
+    def _terminate_bootstrap_process(self):
+        """
+        Terminates the bootstrap subprocess if one is still running.
+        """
+        process = self._bootstrap_process
+        self._bootstrap_process = None
+        terminate_process(process)
```
julien-lang: So the same instance of this object will have multiple Python processes to manage? It's odd.

Could we use a list or Queue instead of a simple variable to store the Python process. So we can better track down and handle edge cases

python/tk_desktop/bootstrap_process.py
```diff
@@ -0,0 +1,54 @@
...
+
+    if process.poll() is not None:
+        return
+
+    logger.debug("Terminating bootstrap process (pid %s)...", process.pid)
+    try:
+        process.terminate()
+    except Exception as exc:
+        logger.warning("Error terminating bootstrap process: %s", exc)
+        return
```
julien-lang: Can you catch a more precised exception?

Also, are we sure we want to return here? I mean, if the process is still alive, we should kill. If not, that means it was a race condition with the previous test and we can safely return

python/tk_desktop/bootstrap_process.py
```diff
@@ -0,0 +1,54 @@
...
+import subprocess
+
+from sgtk import LogManager
+
+logger = LogManager.get_logger(__name__)
+
+DEFAULT_TERMINATION_TIMEOUT = 5
+
+
+def terminate_process(process, timeout=DEFAULT_TERMINATION_TIMEOUT):
```
julien-lang: Please add typing info for all new code

python/tk_desktop/bootstrap_process.py
```diff
@@ -0,0 +1,54 @@
...
+        process.wait(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        logger.warning(
+            "Bootstrap process (pid %s) did not exit, killing...", process.pid
+        )
+        try:
+            process.kill()
+            process.wait(timeout=timeout)
+        except Exception as exc:
+            logger.warning("Error killing bootstrap process: %s", exc)
```
julien-lang:
```suggestion
    try:
        process.wait(timeout=timeout)
        return
    except subprocess.TimeoutExpired:
        logger.warning(
            "Bootstrap process (pid %s) did not exit, killing...", process.pid
        )

     try:
         process.kill()
     except Exception as exc:
         logger.warning("Error killing bootstrap process: %s", exc)
         
      try:
          process.wait(timeout=timeout)
      except subprocess.TimeoutExpired:
          # what do we do now... .we tried our best I guess..... This is why keeping a list might be better. So we re-attempt later....
```

python/tk_desktop/site_communication.py
```diff
@@ -31,6 +32,31 @@ class SiteCommunication(QtCore.QObject, CommunicationBase):
...
+
+    def set_bootstrap_process(self, process):
+        """
+        Registers the bootstrap subprocess and terminates any previous one.
+
+        :param process: A :class:`subprocess.Popen` instance for the bootstrap process.
+        """
+        self._terminate_bootstrap_process()
+        self._bootstrap_process = process
+
```
julien-lang, line 47: I understand that this is the worst case scenario. But where is the best case scenario? I mean, the Python process exits properly and we don't have any remaining process.

julien-lang (commented):
This is intersting. Thanks for the PR.

Are the symptons equivalents on each OS? Windows/Linux/macOS

tests/test_bootstrap_process.py
joel-fortin: I'll update it.

joel-fortin:
> This is intersting. Thanks for the PR.
> 
> Are the symptons equivalents on each OS? Windows/Linux/macOS

The orphan process exists on all 3 OSes, but the cache lock symptom is mostly Windows — macOS/Linux use advisory file locking so cache deletion usually still works there. The subprocess staying alive is universal though. I don't think a code change is needed here.

tests/test_bootstrap_process.py
joel-fortin: Yes can use this convention if you prefer. I took the same pattern as other test files, that's all.

python/tk_desktop/site_communication.py
joel-fortin: SiteCommunication manages exactly one active project at a time — Desktop is not multi-project. set_bootstrap_process() already terminates the old process and stores the new one, so the single variable accurately models the reality. A list would add complexity without solving a real problem imo. Should we stick to a single-project model and keep the single variable ?

python/tk_desktop/bootstrap_process.py
joel-fortin: Yep worth implementing.

python/tk_desktop/bootstrap_process.py
joel-fortin: I'll add it.

python/tk_desktop/bootstrap_process.py
joel-fortin: I'll add these cleaner try/except. Also covering edge cases.

python/tk_desktop/site_communication.py
joel-fortin: shut_down() here is an override of CommunicationBase.shut_down(), which was already being called by the existing code. We didn't add any new callsites — we simply deepened what the method does. The override first calls the base class (RPC disconnect), then terminates the bootstrap subprocess.

It is called from 4 existing locations:

desktop_window.py — _about_to_quit() _(user quits Desktop via menu / Alt+F4)_
desktop_window.py — _on_back_to_projects_clicked() _(user clicks back to the project browser)_
desktop_window.py — __launch_app_proxy_for_project() _(before switching to a new project)_
desktop_engine_site_implementation.py — destroy_engine() (_engine teardown)_

tests/test_bootstrap_process.py
joel-fortin: I could clean up the sys.path. The insert is perform across all the test so I'll need to fix all the tests.

python/tk_desktop/site_communication.py
joel-fortin: In the normal/best case, the shutdown sequence in CommunicationBase.shut_down() (called before _terminate_bootstrap_process()) sends a signal_disconnect RPC call to the bootstrap process. The bootstrap process receives that signal and exits on its own.

By the time _terminate_bootstrap_process() runs, process.poll() already returns a non-None exit code, so terminate_process() detects the process is gone and returns immediately — no terminate() or kill() is ever called.

here the _terminate_bootstrap_process() is a safety net for the cases where the RPC disconnect doesn't work.

tests/test_bootstrap_process.py
julien-lang: No I mean, I never though about using a `with patch.dict` to import the modules. I like it and we might want to generalize it.

Also I am chasing a core swap issue (SG-43592) where I suspect the reason is because we do not clean up the `sys.path` after importing our modules.

hooks/launch_python.py
```diff
@@ -59,7 +59,12 @@ def execute(self, project_python, pickle_data_path, utilities_module_path):
         # will be shared with the child process and it prevent restarting the server
         # after the process closes.
         # Solution was found here: http://stackoverflow.com/a/13593715
-        subprocess.Popen(args, startupinfo=startupinfo, close_fds=True)
+        process = subprocess.Popen(args, startupinfo=startupinfo, close_fds=True)
+
+        # Keep track of the bootstrap process so it can be terminated when the
+        # desktop app disconnects from the project or quits.
+        if hasattr(self.parent, "site_comm"):
+            self.parent.site_comm.set_bootstrap_process(process)
```
julien-lang: More direct and fault tolerant:
```suggestion
        try:
            self.parent.site_comm.set_bootstrap_process(process)
        except AttributeError:
            pass
```

python/tk_desktop/site_communication.py
```diff
@@ -31,6 +32,31 @@ class SiteCommunication(QtCore.QObject, CommunicationBase):
...
+        """
+        process = self._bootstrap_process
+        self._bootstrap_process = None
+        terminate_process(process)
+
+    def shut_down(self):
+        """
+        Disconnects from the project proxy and terminates the bootstrap subprocess.
+        """
+        CommunicationBase.shut_down(self)
```
julien-lang: No more Py2 syntax

```suggestion
        super().shut_down()
```

tests/test_bootstrap_process.py
joel-fortin: check [bdd2647](https://github.com/shotgunsoftware/tk-desktop/pull/221/commits/bdd26472de0f53ee23cdc9f834f5a7870fae0b3f)

tests/test_bootstrap_process.py
joel-fortin: check [bdd2647](https://github.com/shotgunsoftware/tk-desktop/pull/221/commits/bdd26472de0f53ee23cdc9f834f5a7870fae0b3f)

python/tk_desktop/bootstrap_process.py
joel-fortin: check [9f88832](https://github.com/shotgunsoftware/tk-desktop/pull/221/commits/9f8883219c214a9c316346898b2f3a133098649c) and [dea4d2a](https://github.com/shotgunsoftware/tk-desktop/pull/221/commits/dea4d2adabb5203d20d9a17c5ab8428547eac641)

python/tk_desktop/bootstrap_process.py
joel-fortin: check [4a0f7ae](https://github.com/shotgunsoftware/tk-desktop/pull/221/commits/4a0f7ae52cd40e4700c36c9a4763c20462858685)

python/tk_desktop/bootstrap_process.py
joel-fortin: check [4a0f7ae](https://github.com/shotgunsoftware/tk-desktop/pull/221/commits/4a0f7ae52cd40e4700c36c9a4763c20462858685)

hooks/launch_python.py
joel-fortin: check [f5b775c](https://github.com/shotgunsoftware/tk-desktop/pull/221/commits/f5b775c20120380527276af4e6ec8117731d4031)

python/tk_desktop/site_communication.py
joel-fortin: check [8e0fa93](https://github.com/shotgunsoftware/tk-desktop/pull/221/commits/8e0fa938de11105f49c1ede7b5330a526aedaaa6)

python/tk_desktop/site_communication.py
joel-fortin: added docstring for override usage in [dea4d2a](https://github.com/shotgunsoftware/tk-desktop/pull/221/commits/dea4d2adabb5203d20d9a17c5ab8428547eac641)

python/tk_desktop/bootstrap_process.py
```diff
@@ -0,0 +1,66 @@
...
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+Helpers for managing the project bootstrap subprocess launched by tk-desktop.
+"""
+
+import subprocess
+from typing import Optional
```
julien-lang: import typing


...

typing.Optional

python/tk_desktop/bootstrap_process.py
```diff
@@ -0,0 +1,66 @@
...
+        process.kill()
+    except OSError as exc:
+        logger.warning("Error killing bootstrap process: %s", exc)
+
+    try:
+        process.wait(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        logger.warning(
+            "Bootstrap process (pid %s) did not exit after kill().", process.pid
+        )
```
julien-lang, line 66: Ok so this is the worst case scenario here. We don't have any ways to improve the situation....
But this would be a new issue since all the code above will fix the current known issue :D

python/tk_desktop/site_communication.py
```diff
@@ -12,7 +12,10 @@
 Implements communication channels between the desktop app and the background process.
 """
 
+import subprocess
+
 from sgtk.platform.qt import QtCore
+from .bootstrap_process import terminate_process
```
julien-lang: Same here, I would prefer only importing `bootstrap_process` and then calling `bootstrap_process.terminate_process`.

The reason for this preference is to keep it clear what is local or builtin and what is imported.

python/tk_desktop/site_communication.py
```diff
@@ -31,6 +34,38 @@ class SiteCommunication(QtCore.QObject, CommunicationBase):
...
+        bootstrap subprocess after disconnecting from the RPC proxy.
+
+        Called from the following locations:
+        - ``desktop_window._about_to_quit``: user quits Desktop via menu / Alt+F4
+        - ``desktop_window._on_back_to_projects_clicked``: user returns to the project browser
+        - ``desktop_window.__launch_app_proxy_for_project``: before switching to a new project
+        - ``desktop_engine_site_implementation.destroy_engine``: engine teardown
+        """
+        super().shut_down()
+        self._terminate_bootstrap_process()
```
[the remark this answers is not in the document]
julien-lang: LOL sorry about that. I was the one asking about using super. But I forgot this Qt thing... with double inheritance.

julien-lang (approved):
LGTM!

carlos-villavicencio-adsk (approved):

PR #224, opened 2026-06-22
SG-43461 SG-43644 Migrate Flow host support for Desktop engine

yungsiow:
Implement subclass of FlowHost for Desktop engine
Instantiate DesktopHost instance during engine set up with context is configured with flow integration


engine.py
```diff
@@ -150,6 +150,12 @@ def post_app_init(self):
         if hasattr(self.__impl, "post_app_init"):
             self.__impl.post_app_init()
 
+        # Instantiate FlowHost if current context is configured with Flow
+        if self.context.flow_project_id:
+            self.logger.info("Instantiating Flow host as DesktopHost...")
+            host_mod = self.import_module("flowam.host")
+            self._flow_host = host_mod.DesktopHost(self.context)
```
chenm1adsk, line 157: Should we exposing this flow_host as a property of Desktop engine?

python/flowam/host.py
```diff
@@ -0,0 +1,245 @@
...
+        file_type: str = "",
+        multi_select: bool = False,
+    ) -> list[str]:
+        """Invoke a file dialog for selecting one or more file paths.
+
+        Args:
+            title: Title of dialog.
+            starting_dir: Starting location of dialog.
+            folder_mode: If True, dialog will browse folders instead of files.
+            file_filter: Extension of file type to filter for.
```
chenm1adsk: Please update the stale docstring here

python/flowam/host.py
```diff
@@ -0,0 +1,245 @@
...
+
+        # Show dialog
+        dialog.exec()
+
+        # Determine which button is clicked and return corresponding action index
+        # NOTE: from QMessageBox documentation:
+        #       When you add custom buttons with addButton(), the return value of
+        #       exec_() can vary across platforms, but clickedButton() will consistently
+        #       return the actual button object that was clicked.
+        return push_buttons.index(dialog.clickedButton())
```
chenm1adsk, line 143: Copilot said that dialog.clickedButton() returns None when the user dismisses the dialog (closes window) when no cancel provided. And  push_buttons.index(None) will raises ValueError. Maybe need to double check here

documentation: https://doc.qt.io/qt-6/qmessagebox.html#clickedButton

python/flowam/host.py
```diff
@@ -0,0 +1,245 @@
...
+                     NOTE: Maximum of 9 buttons can be supported for this host.
+            default: Index of default button.
+            cancel: Index of cancel button. If not specified, native behaviour is to
+                    use 1 if there is more than one option, or 0 otherwise.
+            no_ui_option: Irrelevant for Desktop host.
+
+        Returns:
+            The index of the button selected by user. Value of -1 indicates dismissed dialog.
+        """
+        from tank.platform.qt import QtGui as qtg
```
carlos-villavicencio-adsk: There's no need to add a simpler alias naming for this. In TK we use `QtGui` as is, even if the name gets longer.

python/flowam/host.py
```diff
@@ -0,0 +1,245 @@
...
+            parent=parent,
+            caption=title,
+            dir=starting_dir,
+            filter=file_filter,
+        )
+        if result:
+            return [str(result)]
+        return []
+
+    def _get_dialog_parent(self):
```
carlos-villavicencio-adsk, line 208: Maybe let's add return annotation?

python/flowam/host.py
```diff
@@ -0,0 +1,245 @@
...
+    def copy_to_clipboard(self, text: str) -> bool:
+        """Copy given text to clipboard of QT application.
+
+        Args:
+            text: Text to be copied.
+
+        Returns:
+            True on success.
+        """
+        from tank.platform.qt import QtGui as qtg
```
carlos-villavicencio-adsk: Same comment as before.

carlos-villavicencio-adsk (commented):
Looks good so far. I've pointed out some suggestions.

engine.py
yungsiow: it is exposed as part of engine class - see tk-core PR.  There is a "flow_host" property that returns this "private" variable.  If I'm understanding your comment correctly!

python/flowam/host.py
yungsiow: good catch!

python/flowam/host.py
yungsiow: Let me test that!

python/flowam/host.py
yungsiow: I think this will be tricky because the return type is a QWidget, and I don't think we want to add top-level qt imports in this module.  Can I just add that return type to the docstring instead?

python/flowam/host.py
yungsiow: Hey so I tried it via script, and didn't run into any errors.  The return value was 1 when I didn't specify a "cancel" option.  I think I'll leave this for now.  If there's a future issue, let's address it then, but I can't seem to reproduce this scenario.

yungsiow: Ah I see why, we always set a cancel value internally if it was not set by the caller.  So we'll never run into the scenario above.  I think we're safe :)

chenm1adsk: Nice! Thank you for double confirming!

chenm1adsk (approved):
LGTM!

carlos-villavicencio-adsk (approved):
LGTM

python/flowam/host.py
yungsiow: fixed!

python/flowam/host.py
yungsiow: fixed!

PR #229, opened 2026-09-18
SG-43379 Fix missing desktop menu action icons

carlos-villavicencio-adsk:
This pull request addresses an issue where context-menu action icons were missing in FPTR Desktop v3.0 on macOS, due to Qt's default behavior of hiding icons in menus. The main change ensures icons are always visible in context menus by explicitly setting their visibility, and comprehensive unit tests are added to verify this fix.

image


python/tk_desktop/desktop_engine_site_implementation.py
```diff
@@ -280,6 +280,11 @@ def trigger_register_command(self, name, properties, groups):
             action = QtGui.QAction(self.desktop_window)
             if icon is not None:
                 action.setIcon(icon)
+                # On macOS, Qt sets AA_DontShowIconsInMenus to True by default,
+                # which suppresses icons in all QMenu items. Override this per
+                # action so icons remain visible regardless of the host
+                # application's attribute setting (changed in FPTR Desktop v3.0).
+                action.setIconVisibleInMenu(True)
```
julien-lang, line 287: Ok but why was it working in SGD 2?

carlos-villavicencio-adsk: The only change that comes to my mind is the change from pyside 6.5 commercial to 6.8 LGPL.

stevelittlefish (approved):

